PR #10, opened 2024-10-18
feat(nova): enabled call page to have fullscreen

mostafakamar2308:


neuodev (commented):
The code looks good to me! Great work @mostafakamar2308! 

I don't understand why the branch `mk/call-fullscreen` is based on `mk/call-feedback`. It has nothing to do with `mk/full-feedback`. 

I suggest rebasing `mk/call-fullscreen` on top of `master` then change the base of the pull request to `master` as well.

apps/nova/src/hooks/call.ts
```diff
@@ -357,3 +357,47 @@ export function useCallEvents(
...
   };
 }
+
+export function useFullScreen() {
+  const [isFullScreen, setIsFullScreen] = useState(false);
+
+  const startFullScreen = useCallback(async () => {
+    try {
+      setIsFullScreen(true);
+      await document.body.requestFullscreen();
```
neuodev, line 367: We don't want to whole body to be full screen. We want this [`div`](https://github.com/litespace-org/litespace/blob/26f7e8691d90effdd17d3d4b8210949cdb71e071/apps/nova/src/pages/Call.tsx#L189)  to be full screen. You can easily achieve this by using react refs. 

### Example

```ts

function useFullScreen() {
   const ref = useRef<HTMLDivElement>(null);

  const toggle = useCallback(() => {
    if (!ref.current) return;
    ref.current.requestFullScreen();
  }, [)


  return {ref, toggle, isFullScreen}
}


// in Call.tsx

const {ref, toggle} = useFullScreen()

return <div ref={ref}> </div>

```

apps/nova/src/hooks/call.ts
```diff
@@ -357,3 +357,47 @@ export function useCallEvents(
...
+    } catch (error) {
+      console.log(error);
```
neuodev, line 369: Please use `toaster.error` to show error notification in case the request failed.

neuodev (changes requested):

apps/nova/src/hooks/call.ts
```diff
@@ -357,3 +357,56 @@ export function useCallEvents(
...
+      if (ref.current) {
+        setIsFullScreen(true);
+        await ref.current.requestFullscreen();
+      }
+    } catch (error) {
+      console.log(error);
+      toaster.error({
+        title: intl("error.unexpected"),
+      });
+    }
```
neuodev, line 377:
```suggestion
   if (!ref.current) return 
    try {
        await ref.current.requestFullscreen();
    } catch (error) {
      toaster.error({
        title: intl("error.unexpected"),
      });
    }
```

It is better to se the `isFullScreen` using the `fullscreenchange` event. Here is an [example](https://github.com/litespace-org/litespace/blob/c45fc5f29549d49a4ede2a9b668f2cebe8bf25c7/packages/luna/src/components/VideoPlayer/video.ts#L179-L189) 

```ts
  const onFullScreen = useCallback(() => {
    // check if the `fullScreenElement` is the same as `ref` (example: compare element ids)
    const fullscreen = document.fullscreenElement == ref.current; // <--- sudo code 
    setFullScreen(fullscreen)
  }, []);

  useEffect(() => {
    document.addEventListener("fullscreenchange", onFullScreen);
    return () => {
      document.removeEventListener("fullscreenchange", onFullScreen);
    };
  }, [onFullScreen]);
```

apps/nova/src/hooks/call.ts
```diff
@@ -357,3 +357,56 @@ export function useCallEvents(
     mateVideo,
   };
 }
+
+export function useFullScreen() {
+  const intl = useFormatMessage();
+  const ref = useRef<HTMLDivElement>(null);
+  const [isFullScreen, setIsFullScreen] = useState(false);
```
neuodev, line 364:
```suggestion
  const [fullScreen, setFullScreen] = useState<boolean>(false);
```

neuodev (changes requested):

apps/nova/src/hooks/call.ts
```diff
@@ -357,3 +357,60 @@ export function useCallEvents(
...
+      toaster.error({
+        title: intl("error.unexpected"),
+      });
+    }
+  }, []);
+
+  const exitFullscreen = useCallback(async () => {
+    if (ref.current) {
+      await document.exitFullscreen();
+    }
```
neuodev, line 385:
```suggestion
    if (ref.current) await document.exitFullscreen();
```

apps/nova/src/hooks/call.ts
```diff
@@ -357,3 +357,60 @@ export function useCallEvents(
...
+  }, []);
+
+  const toggleFullScreen = useCallback(async () => {
+    if (!document.fullscreenElement) {
+      startFullScreen();
+    } else {
+      exitFullscreen();
+    }
```
neuodev, line 393: It can be cleaner by doing this. 

```suggestion
  const toggleFullScreen = useCallback(async () => {
    if (!document.fullscreenElement) return startFullScreen();
    return exitFullscreen();
```

apps/nova/src/hooks/call.ts
```diff
@@ -357,3 +357,60 @@ export function useCallEvents(
...
+
+  const toggleFullScreenByKeyboard = useCallback((e: KeyboardEvent) => {
+    if (e.code === "F11") {
+      // I want to control full screen myself to expose the state for UI purposes
+      e.preventDefault();
+      toggleFullScreen();
+    }
+    if (e.code === "Escape") {
+      exitFullscreen();
+    }
```
neuodev, line 404: Cleaner. 

```suggestion
    e.preventDefault();
    if (e.code === "F11") return toggleFullScreen();
    if (e.code === "Escape") return exitFullscreen();
```

apps/nova/src/pages/Call.tsx
```diff
@@ -190,6 +195,7 @@ const Call: React.FC = () => {
   return (
     <div className="flex h-[calc(100vh-4rem)] overflow-hidden w-full">
       <div
+        ref={ref}
```
neuodev, line 198: The `ref` must have an ID to be able to compare it to the `document.fullscreenElement.id`
```suggestion
      <div
        ref={ref}
        id="call-page"
```

apps/nova/src/hooks/call.ts
```diff
@@ -357,3 +357,60 @@ export function useCallEvents(
...
+
+  const startFullScreen = useCallback(async () => {
+    try {
+      if (!ref.current) return;
+      await ref.current.requestFullscreen();
+    } catch (error) {
+      toaster.error({
+        title: intl("error.unexpected"),
+      });
+    }
```
neuodev, line 379: Use our `safe` util from `@litespace/sol`

```suggestion
  const result = await safe(async () => {
      if (!ref.current) return;
      await ref.current.requestFullscreen();
   })
   if (result instanceof Error) toaster.error({ title: intl("error.unexpected") });
```

neuodev (changes requested):

apps/nova/src/hooks/call.ts
```diff
@@ -357,3 +357,48 @@ export function useCallEvents(
     mateVideo,
   };
 }
+
+export function useFullScreen() {
+  const intl = useFormatMessage();
+  const ref = useRef<HTMLDivElement>(null);
+  const [isFullScreen, setIsFullScreen] = useState<boolean>(false);
```
neuodev, line 364:
```suggestion
  const [fullScreen, setFullScreen] = useState<boolean>(false);
```

neuodev (approved):

PR #38, opened 2024-11-02
refactor(luna): using ts instead of rollup to bundle into different files

mostafakamar2308:


packages/luna/tsconfig.json
```diff
@@ -1,27 +1,39 @@
...
+    "noFallthroughCasesInSwitch": true,
+    "useDefineForClassFields": true,
+    "skipLibCheck": true,
+    "allowSyntheticDefaultImports": true,
+    "forceConsistentCasingInFileNames": true,
+    "isolatedModules": true
   },
-  "include": ["src"],
-  "references": [{ "path": "./tsconfig.node.json" }]
+  "include": ["src", ".storybook", "src/**/*.stories.tsx"]
```
neuodev, line 38: This way we will bundle storybooks along side our components which something we don't want to do. 

To resolve this issue you should do the following. 

In `tsconfig.json`  keep `include` as it is. 

```json
  "include": ["src", ".storybook", "src/**/*.stories.tsx"]
```
Create another config file and call it `tsconfig.build.json`. It should extend `tsconfig.json`. It should include the `src/` only. This way we will only include the components in our build. 

```ts
{
  "extends": "./tsconfig.json",
  "include": ["src/"], 
  "exclude": ["src/**/*.stories.txs"],
}
```

When building the components make sure to use `tsconfig.build.json` and not the default `tsconfig.json`.

neuodev (changes requested):

packages/luna/package.json
```diff
@@ -6,38 +6,83 @@
...
   ],
   "scripts": {
     "dev": "vite",
     "clean": "rm -rf dist/",
-    "build": "tsc && vite build && pnpm build:css",
+    "build": "tsc -p tsconfig.json && tsc-alias -p tsconfig.json && pnpm run build:css",
     "watch:pkg": "vite build -w",
     "watch:build": "concurrently --names \"pkg,css\" \"pnpm watch:pkg\" \"pnpm build:css -w\"",
-    "build:css": "tailwindcss -m -i src/tailwind.css -o dist/style.css",
+    "build:css": "tailwindcss -m -i src/tailwind.css -o dist/style.css && copyfiles --flat src/tailwind.css tailwind.config.js dist/",
```
neuodev, line 85: No need for this, TypeScript will copy the file to dist. 

```suggestion
    "build:css": "tailwindcss -m -i src/tailwind.css -o dist/style.css",
```

mostafakamar2308: It didn't do that, even though they were used in `package.json`
```
"./tailwind.config": "./dist/tailwind.config.js",
```

neuodev:
🚀 Nova: https://nova-k5473xrtt-litespace-cores-projects.vercel.app
🚀 Luna: https://luna-hxxyjj8tn-litespace-cores-projects.vercel.app
🚀 Nebula: https://nebula-apw1cgsns-litespace-cores-projects.vercel.app

PR #50, opened 2024-11-04
refactor(luna): update toast component to use radix

moalidv:


packages/luna/src/components/Toast/RadixToast.tsx
```diff
@@ -0,0 +1,77 @@
...
+  setType: (type: ToastType) => void;
+  setOpen: (open: boolean) => void;
+  setTitle: (title: string) => void;
+};
+
+export const ToastContext = createContext<ToastContextTypes | null>(null);
+
+export const ToastProvider = ({ children }: ToastContextProps) => {
+  const [open, setOpen] = useState(false);
+  const [title, setTitle] = useState("");
```
mostafakamar2308, line 24: We would Also need description state here, like for error descriptions

packages/luna/src/components/Toast/RadixToast.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const [type, setType] = useState<ToastType>("success");
+  return (
+    <ToastContext.Provider
+      value={{ open, type, title, setOpen, setTitle, setType }}
```
mostafakamar2308, line 28: Instead of Exposing all of these values, Let's think of this differently.
Every Component will need to just open a toast. 
So how about we make a function that opens a toast and can accept params like `type, title, description`.
Then This will be exposed afterward in other components
like so:
```
const showToaster = (type, title, description?)=>{
setOpen(true)
setTitle(title)
setDescription(description)
}
```

Then we can use it in different components like:
```
const toaster = useToast();
toaster.showToast("Success", "Success Title", "success long description")
```

packages/luna/src/components/Toast/RadixToast.tsx
```diff
@@ -0,0 +1,77 @@
+import { createContext, useContext, useState } from "react";
+import * as Toast from "@radix-ui/react-toast";
+import cn from "classnames";
+
+type ToastType = "success" | "failed";
```
mostafakamar2308, line 5: Also How about if you also exported this as an enum, as this is a general component that can be used everywhere. so if we added or changed one of the types, then you won't need to change it everywhere it's used again.
You can look at this [example](https://github.com/litespace-org/litespace/blob/e7e13b8b7ac2288aedd57fef5a27cf41f7371f4e/packages/luna/src/components/Button/types.ts#L1)

mostafakamar2308 (changes requested):

neuodev:
Close in favor of #55

PR #124, opened 2024-11-21
feat(ui): impl.  user context provider

neuodev:


neuodev:
Closed in favor of #182

PR #170, opened 2024-12-05
added: cache/call.ts full radis implementaion.

mmoehabb:
- The call class (radis controller) has been implemented.
- radis data seeded (for only call table),
- server handlers now works with radis,
- server tests pass,
- and initial implementation for frontend hooks appertain to call members added; radis server synchronously reflects actual members in calls (on each user join and leave).


neuodev:
Will be closed in favor of #173

PR #206, opened 2024-12-12
added: initial files for database ERD.

mmoehabb:


neuodev:
As discussed internally, the plan will be to host the diagram on DrawDB and just add the link of it in the docs.

PR #239, opened 2024-12-22
added: flushing the cache on server startup.

mmoehabb:


services/server/src/index.ts
```diff
@@ -20,6 +20,10 @@ import "colors";
 // Stablish connection with the redis cache.
 cache.connect();
 
+// Flush the cache as the data in cache might not be compatible with the changes 
+// in the server itself which might lead to undefined behavior. 
+cache.flush();
```
neuodev, line 25: I think we should flush the when the connection is established. 

```
cache.connect().then(() => {
 cache.flush();
})
```

neuodev (changes requested):

neuodev (approved):

PR #242, opened 2024-12-23
added: coverage unit tests for redis cache component.

mmoehabb:


packages/models/tests/cache/call.test.ts
```diff
@@ -0,0 +1,64 @@
+import { Cache } from "@/cache";
+import { expect } from "chai";
+
+const cache = new Cache(process.env.REDIS_URL || "");
```
neuodev, line 4: No to create this cache object for each test file, we can move it into a fixture.

neuodev (changes requested):

mmoehabb:
The pipeline before [this change](https://github.com/litespace-org/litespace/pull/242/commits/9ce9716f49058ad08dffb0994308db2f6fe9a887) kept raising an error due to primary key violation. Local testing, on the other hand, raises no errors at all. I've worked around this tedious error by using a hard-coded values in the corresponding unit test; as shown in [the commit](https://github.com/litespace-org/litespace/pull/242/commits/9ce9716f49058ad08dffb0994308db2f6fe9a887).

neuodev:
> The pipeline before [this change](https://github.com/litespace-org/litespace/pull/242/commits/9ce9716f49058ad08dffb0994308db2f6fe9a887) kept raising an error due to primary key violation. Local testing, on the other hand, raises no errors at all. I've worked around this tedious error by using a hard-coded values in the corresponding unit test; as shown in [the commit](https://github.com/litespace-org/litespace/pull/242/commits/9ce9716f49058ad08dffb0994308db2f6fe9a887).

Sounds good 👌

neuodev (approved):

PR #243, opened 2024-12-23
added: seeding data for topics table

mmoehabb:


packages/models/scripts/seed.ts
```diff
@@ -256,6 +256,29 @@ async function main(): Promise<void> {
...
+        mockTopics[i] = "_" + mockTopics[i] + "_";
+      }
+    }
+
+    return await Promise.all(
+      mockTopics.map(async (topic: string) => {
+        await topics.create({
+          name: {
+            ar: topic,
+            en: topic,
```
neuodev, line 275: make sure that their are tow versions of each topic one is Arabic and one is English.

packages/models/scripts/seed.ts
```diff
@@ -256,6 +256,29 @@ async function main(): Promise<void> {
...
+  // seeding topics data
+  await knex.transaction(async (tx) => {
+    stdout.info(`Inserting 50 random topics in the database.`);
+
+    const mockTopics = faker.lorem.words(50).split(" ");
+    for (let i = 0; i < mockTopics.length; i++) {
+      if (mockTopics.slice(i+1, mockTopics.length).includes(mockTopics[i])) {
+        mockTopics[i] = "_" + mockTopics[i] + "_";
+      }
+    }
```
neuodev, line 268: No need to do this. It is not that clean. Also it doesn't create English and Arabic topics. 

```ts
range(50).forEach(() => {
  const ar = fakerAr.lorem.word();
  const en = fakerEn.lorem.word();
  await topics.create({name: {ar, en}}, tx); 
})
```

neuodev (changes requested):

neuodev (approved):

PR #254, opened 2024-12-26
refactor(luna): impl loading and error in lessons page

moalidv:


neuodev:
🚀 [Nova](https://nova-9e9d6t62v-litespace-core.vercel.app) 
🚀 [Luna](https://luna-68w9bcwbc-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-rf5shvkwi-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-c8e2ru2sb-litespace-core.vercel.app)

packages/assets/assets/loading.svg
mostafakamar2308, line 1: I think it would be better when you aren't using the svg, to not include it

packages/models/tests/ratings.test.ts
```diff
@@ -79,7 +79,9 @@ describe("Ratings", () => {
         values: [1, 2, 4, 5],
       });
 
-      const avg = await ratings.findAvgRatings([t1.id, t2.id, t3.id]);
+      const avg = await ratings.findAvgRatings({
```
mostafakamar2308, line 82: unrelated changes, Try to rebase on latest master

moalidv: image

@mostafakamar2308 
I want to ask about that point 
findAvgRatings function take an object as a parameter containing users as an array of numbers so how can we give the function this array directly without the object and label of users

moalidv: @mostafakamar2308 
I have tried to rebase on the latest master but the error is still present

neuodev:
@moalidv please rebase.

packages/luna/src/components/Loading/Loader.tsx
```diff
@@ -19,32 +19,36 @@ export const Loader: React.FC<{
...
-            variant === "large"
-              ? "radial-gradient(circle, transparent 30px, white 16px)"
-              : "radial-gradient(circle, transparent 25px, white 0px)",
+          maskImage: `radial-gradient(circle, transparent 
+          ${variant === "large" ? "30px" : ""}
+				  ${variant === "medium" ? "25px" : ""}
+				  ${variant === "small" ? "15px" : ""}, white
+				  ${variant === "large" ? "16px" : ""}
+				  ${variant === "medium" ? "0px" : ""}
+          ${variant === "small" ? "0px" : ""}
```
neuodev, line 28: we can have a function that returns the size based on the variant. 

```
const size = useMemo(() => {
 if (variant === 'large') return '30px';
// ... 
}, [])

// or even better 

```ts
const styles = useMemo(() => {
   if (variant === 'large') return ['30px', '16px'];
   // ... 
}, [])
```

packages/assets/assets/loading.svg
neuodev, line 1: This is not in use. Please remove it as well as its export in the `package.json` file.

neuodev (changes requested):

packages/models/tests/ratings.test.ts
neuodev: If you rebased now, you should not face this issue.

neuodev:
There is a degree of distortion in the spinner. 

[Screencast from 2024-12-30 20-06-47.webm](https://github.com/user-attachments/assets/68c5bd77-bb03-4532-b5fd-4a7f657beeda)

apps/nova/src/components/UpcomingLessons/Content.tsx
```diff
@@ -25,8 +25,9 @@ export const Content: React.FC<{
   fetching: boolean;
   error: boolean;
   hasMore: boolean;
+  refetch?: Void;
```
mmoehabb, line 28: I think using the `Void` type, in general, is considered bad or harmful practice. If it's not feasible or not that necessary and tedious to declare/define a type for this specific function. I'd recommend leaving a comment stating what it's supposed to do.

mmoehabb: You may write, for instance: `// A callback function that invoked tanstack query refetch method`.

packages/luna/src/components/Loading/LoadingError.tsx
```diff
@@ -4,15 +4,24 @@ import { useFormatMessage } from "@/hooks/intl";
...
+import cn from "classnames";
 
-export const LoadingError: React.FC<{ retry: Void; error: string }> = ({
-  retry,
-  error,
-}) => {
+export const LoadingError: React.FC<{
+  retry: Void;
+  error: string;
+  variant?: "small" | "medium" | "large";
```
mmoehabb, line 12: Awesome work. This solution is really good and flawless. However, I have a suggestion: what if we defined the type of `variant` as follows `"10" | "16" | "20"`, and then embedded it in tailwind class name `tw-w-${variant}`. I think It could make the code more succinct.

mmoehabb (changes requested):
Easy to read and clean code, keep the good work up :smile:

packages/luna/src/components/Loading/LoadingError.tsx
neuodev: This will not work due to how tailwind scan for classnames.

mmoehabb: > This will not work due to how tailwind scan for classnames.

I think It will work fine, at least inside the `cn` utility function.

neuodev:
This pull request will be closed in favor of #271

PR #267, opened 2024-12-31
feat(nova): impl manage schedule dialog for tutors

moalidv:


neuodev:
🚀 [Nova](https://nova-83otj8ksa-litespace-core.vercel.app) 
🚀 [Luna](https://luna-m1ep2qrfi-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-24i2ul5k8-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-5zh8075vy-litespace-core.vercel.app)

neuodev:
🚀 [Nova](https://nova-fhqld6iuo-litespace-core.vercel.app) 
🚀 [Luna](https://luna-aq4ceslxl-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-7mxjg1dtu-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-avi6wjah8-litespace-core.vercel.app)

neuodev:
🚀 [Nova](https://nova-tr6bdf1hz-litespace-core.vercel.app) 
🚀 [Luna](https://luna-nw5s0zsck-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-3dny8asmn-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-gy7twhpip-litespace-core.vercel.app)

neuodev:
🚀 [Nova](https://nova-bg1lo2szf-litespace-core.vercel.app) 
🚀 [Luna](https://luna-mp4mf8va5-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-71v7owf90-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-b6f16vlc8-litespace-core.vercel.app)

neuodev:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true">Warnings</th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>No clickup task found in this pull request</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f015c1decb482993513d854d8958c22ee41be48e
</p>

neuodev:
🚀 [Nova](https://nova-87gyr7iwi-litespace-core.vercel.app) 
🚀 [Luna](https://luna-8e54xkyr2-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-k2ru4y6lo-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-gfinjmxqu-litespace-core.vercel.app)

packages/luna/src/components/ManageSchedule/ManageSchedule.tsx
```diff
@@ -0,0 +1,383 @@
...
+                        <SlotRow
+                          key={slot.id}
+                          start={orUndefined(slot.start)}
+                          end={orUndefined(slot.end)}
+                          day={iso}
+                          setStart={(value) =>
+                            setSlotDate(slot.id, "start", value)
+                          }
+                          setEnd={(value) => setSlotDate(slot.id, "end", value)}
+                          add={canAdd ? () => addSlot({ day: iso }) : undefined}
```
mmoehabb, line 318: Let's always show the add button if it's the last row to match the prototype. We can do so by changing the false clause.
```TS
add={canAdd ? () => addSlot({ day: iso }) : (isLast ? () => {}: undefined)}
```

packages/luna/src/components/ManageSchedule/ManageSchedule.tsx
```diff
@@ -0,0 +1,383 @@
...
+                          remove={
+                            canRemove ? () => removeSlot(slot.id) : undefined
+                          }
+                        />
+                      );
+                    })}
+
+                    {isEmpty(slots) ? (
+                      <SlotRow
```
mmoehabb, line 327: We may add `add={() => {}}` here in order to show the add button as mentioned in the previous comment.
> Don't forget to margin it properly (we may disable the remove button by disappearing  it).

packages/luna/src/components/ManageSchedule/ManageSchedule.tsx
mmoehabb, line 1: Overall this component is really impressive... great work. However, I think it needs refactoring; It's a little buggy and not that smooth. that's probably because of the repetitive cumbersome computation for the `days` variable in the line 170. 

One feasible solution that came to my mind is to segregate the `div` in line 302 (that contains SlotRows) from the rest of this component, we may call the segregated component "SlotRowsContainer", it shall handle the state of its seven slots itself. Therefore, relieving the burden of managing so many slot states from the parent (ManageSchedule), and distributing it on seven different components.

Last thing to remark, consider dividing this script into several ones; meanly, extract SlotRow into its own ts file.

packages/luna/src/components/ManageSchedule/ManageSchedule.tsx
```diff
@@ -0,0 +1,383 @@
...
+import Close from "@litespace/assets/Close";
+import AddCircle from "@litespace/assets/AddCircle";
+import { orUndefined } from "@litespace/sol/utils";
+import { Button, ButtonVariant } from "@/components/Button";
+import { Loader, LoadingError } from "@/components/Loading";
+import cn from "classnames";
+
+const WEEK_DAYS = 7;
+
+function asOptions(start: Dayjs, end: Dayjs): SelectList<string> {
```
mmoehabb, line 22: All is done, except excluding the selected options from the next SlotRow options. e.g. a user is assigning a new slot from 01:00 PM Saturday to 01:30 PM Saturday, when he/she tries to assign another slot in the same day from 01:00 PM and to 01:30 PM shall not be available.

I believe we can approach this by defining a new parameter here that will be used to exclude some options.
```TS
function asOptions(start: Dayjs, end: Dayjs, exclude: IAvailabilitySlot.Slot[]): SelectList<string> {
...
}
```

> You may use [subtractSlots function](https://github.com/litespace-org/litespace/blob/master/packages/sol/src/availabilitySlots.ts#L110) from `packages/sol`. It can be helpful.

packages/luna/src/components/Checkbox/Checkbox.tsx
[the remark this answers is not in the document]
moalidv: the design of the schedule first contained checkbox 
so I used checkbox and when the check box label was only string so you could only add string and couldn't style it 
so if we make it available for react nodes, we can add typography and style it.

packages/luna/src/components/Loading/LoadingError.tsx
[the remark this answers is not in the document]
moalidv: not undefined we put default value for it.
I can imagine if we frequantly use for example medium sized loading error we can assign the dafault size medium and not write it every time 
like button component, it is by default have variant, type and size

neuodev:
🚀 [Nova](https://nova-l2ro4zscu-litespace-core.vercel.app) 
🚀 [Luna](https://luna-j2r3e8lod-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-plir42oe9-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-hd7wxhh1h-litespace-core.vercel.app)

packages/luna/src/locales/ar-eg.json
```diff
@@ -623,6 +628,9 @@
   "labels.share": "مشاركة",
   "labels.go-back": "الرجوع",
   "labels.coming-soon": "قريبا 🎉",
+  "labels.cancel": "إلغاء",
```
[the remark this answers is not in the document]
neuodev: We are moving away from using "global...." and instead using "labels." directly. 

cc @moalidv @mmoehabb

packages/luna/src/components/ManageSchedule/ManageSchedule.tsx
moalidv: @mmoehabb 
but what if the use add a new line before fill the empty slots

packages/luna/src/components/Select/Select.tsx
```diff
@@ -70,24 +76,28 @@ export const Select = <T extends string | number>({
...
-        <Icon>
-          <ArrowDown
-            data-open={open}
-            className={cn(
-              "tw-justify-self-end",
-              "data-[open=true]:tw-rotate-180 tw-transition-all tw-duration-300"
-            )}
-          />
-        </Icon>
+        {showDropdownIcon ? (
```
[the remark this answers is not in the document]
moalidv: I used to write in that way  (condition && component)  but Ahmed told me to use 
condition ? component : null    pattern    
so for consistancy I started to write in this way.

packages/luna/src/components/Select/Select.tsx
```diff
@@ -70,24 +76,28 @@ export const Select = <T extends string | number>({
...
       >
-        <Value>
-          <Typography
-            className={cn(
-              lable ? "tw-text-natural-950" : "tw-text-natural-400"
-            )}
-          >
-            {lable || placeholder}
-          </Typography>
+        <Value
```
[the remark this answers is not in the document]
moalidv: Ok, but for me to understand more this case 
what if the label is undefined so nothing will be rendered 
what is the diff bet putting the condition and not

neuodev:
🚀 [Nova](https://nova-ho828zrv0-litespace-core.vercel.app) 
🚀 [Luna](https://luna-ggefr556r-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-11mbzaxc7-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-6jn4l6lho-litespace-core.vercel.app)

packages/luna/src/components/ManageSchedule/ManageSchedule.tsx
```diff
@@ -0,0 +1,192 @@
...
+          onClick={close}
+          variant={ButtonVariant.Secondary}
+          className="tw-grow tw-basis-1/2"
+        >
+          <Typography
+            element="body"
+            weight="semibold"
+            className="text-brand-700"
+          >
+            {intl("global.labels.cancel")}
```
mmoehabb, line 173: Sorry for the burden... @neuodev has mentioned that we are moving away from using "global..." to using "labels". Revert this to the initial implementation "labels.cancel".

packages/luna/src/locales/ar-eg.json
```diff
@@ -667,6 +672,8 @@
   "global.schedule.rule.format.onDay": "يوم",
   "global.schedule.rule.format.day": "يوم",
   "global.schedule.rule.format.weekday.seperator": "و",
+  "global.placeholder.from": "من",
```
mmoehabb, line 675: Sorry, let's go back to "placeholder.from". As @neuodev has clarified the new convention.

packages/luna/src/components/Select/Select.tsx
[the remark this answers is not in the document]
mmoehabb: Honestly, I have no idea :face_with_diagonal_mouth:
When I pondered about that. I have guessed that maybe using `label` as the text content (and/or children component) of the `Typography` component isn't "enough" or "legitimate", so to speak, for react to re-render the component whenever `label` changes. It could be a bad implementation of `Typography` or a bug in react. In either case, using `label` outside (but linked to) Typography will do the job.

packages/luna/src/components/Select/Select.tsx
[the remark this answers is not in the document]
mmoehabb: It's just a matter of taste. Let's stick with the convention for consistency.

packages/luna/src/components/Loading/LoadingError.tsx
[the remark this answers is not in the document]
mmoehabb: Aha got it, we added the question mark so the user doesn't have to pass a value and the default is enabled instead.

packages/luna/src/components/ManageSchedule/ManageSchedule.tsx
mmoehabb: Hmm... good point. We need to share this with Zeyad. The current implementation doesn't fully match the prototype. If it's okay, from UI/UX perspective, then let's leave it as it is. If it's not (from my point of view, there is a lot of empty space in the left side of the component), I propose adding disable version of the add button, or easily using a dynamic (smooth) width length rather than static.

packages/luna/src/components/ManageSchedule/SlotRow.tsx
```diff
@@ -0,0 +1,131 @@
...
+    return asOptions(startOfSelection, endOfSelection);
+  }, [day, endOfOptions]);
+
+  const toOptions = useMemo(() => {
+    const startOfSelection = startOfOptions
+      ? dayjs(startOfOptions)
+      : dayjs(day);
+    const endOfSelection = dayjs(day).endOf("day");
+
+    return asOptions(startOfSelection, endOfSelection);
```
mmoehabb, line 63: This should be 
```TS
return asOptions(startOfSelection.add(30, "minutes"), endOfSelection);
````
in order to exclude the selected time in `from` from the options in `to`.

File: packages/luna/src/components/ManageSchedule/SlotRow.tsx [hunk 10]
```diff
@@ -0,0 +1,131 @@
...
+    },
+    [onToChange]
+  );
+
+  return (
+    <div className="tw-flex tw-items-center tw-gap-6 tw-w-[400px]">
+      <div className="tw-flex tw-items-center tw-justify-center tw-gap-4">
+        <div className="tw-w-[136px]">
+          <Select
+            value={start}
```
mmoehabb, line 89: Use here `startOfOptions` rather than `start`.

packages/luna/src/components/ManageSchedule/SlotRow.tsx
```diff
@@ -0,0 +1,131 @@
...
+        <Typography
+          element="body"
+          weight="bold"
+          className="tw-text-natural-500"
+        >
+          -
+        </Typography>
+        <div className="tw-w-[136px]">
+          <Select
+            value={end}
```
mmoehabb, line 105: Use here `endOfOptions` rather than `end`.

packages/luna/src/components/ManageSchedule/SlotRow.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+      <div className="tw-flex tw-gap-4">
+        {remove ? (
+          <button type="button" onClick={remove}>
+            <Close />
+          </button>
+        ) : null}
+
+        {add ? (
+          <button type="button" onClick={add}>
```
mmoehabb, line 122: Replace this line with the following:
```TS
<button type="button" onClick={() => add(startOfOptions, endOfOptions)}>
```
> See my comments on `SlotRowContainer.tsx` first

packages/luna/src/components/ManageSchedule/SlotRow.tsx
```diff
@@ -0,0 +1,131 @@
...
+  return options;
+}
+
+const SlotRow: React.FC<{
+  day?: string;
+  start?: string;
+  end?: string;
+  remove?: Void;
+  add?: Void;
+  onFromChange?: (value: string) => void;
```
mmoehabb, line 38: We can remove these two: `onFromChange` and `onToChange`.
> See comments on `SlotRowContainer.tsx` before this one.

packages/luna/src/components/ManageSchedule/SlotRowContainer.tsx
```diff
@@ -0,0 +1,90 @@
...
+        const cloned = structuredClone(prev);
+        const filtered = cloned.filter((slot) => slot.id !== id);
+
+        return filtered;
+      });
+    },
+    [setSlots]
+  );
+
+  const setSlotDate = useCallback(
```
mmoehabb, line 46: Modify this function to set both dates rather than one at a time.
```TS
const setSlotDates = useCallback(
  (id: number, start: string, end: string) => {
    setSlots((prev) => {
      const cloned = structuredClone(prev);
      const foundSlot = cloned.find((slot) => slot.id === id);
      if (foundSlot) { 
        foundSlot.start = start; 
        foundSlot.end = end;
      }
      return cloned;
    });
  },
  [setSlots]
);
```

packages/luna/src/components/ManageSchedule/SlotRowContainer.tsx
```diff
@@ -0,0 +1,90 @@
...
+        const canAdd = isLast && slot.start && slot.end;
+        const canRemove = slot.start && slot.end;
+
+        return (
+          <SlotRow
+            key={slot.id}
+            start={orUndefined(slot.start)}
+            end={orUndefined(slot.end)}
+            day={iso}
+            onFromChange={(value) => setSlotDate(slot.id, "start", value)}
```
mmoehabb, line 71: We should remove those two: `onFromChange` and `onToChange`; rendering slot rows each time a user selects a time is cumbersome.

packages/luna/src/components/ManageSchedule/SlotRowContainer.tsx
```diff
@@ -0,0 +1,90 @@
...
+            onToChange={(value) => setSlotDate(slot.id, "end", value)}
+            add={canAdd ? () => addSlot({ day: iso }) : undefined}
```
mmoehabb, line 73: Change this to the following:
```TS
add={isLast ? (start, end) => {
  setSlotDates(slot.id, start, end)
  addSlot({ day: iso })
} : undefined}
```

packages/luna/src/components/ManageSchedule/SlotRowContainer.tsx
```diff
@@ -0,0 +1,90 @@
...
+            remove={canRemove ? () => removeSlot(slot.id) : undefined}
+          />
+        );
+      })}
+
+      {isEmpty(slots) ? (
+        <SlotRow
+          day={iso}
+          onFromChange={(value: string) => addSlot({ day: iso, start: value })}
```
mmoehabb, line 82: And finally, replace those two with the following:
```TS
add={(start, end) => {
  addSlot({ day: iso, start, end })
  addSlot({ day: iso })
}}
```

mmoehabb (changes requested):
All looks good. However, the component still needs some enhancements. I have added my suggestion to enhance it's rendering a little; by invoking re-rendering only when the user adds (and not selecting) slots.

neuodev:
🚀 [Nova](https://nova-fmw1f6ct3-litespace-core.vercel.app) 
🚀 [Luna](https://luna-km62qnd76-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-22mbbeq13-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-a3qec4np9-litespace-core.vercel.app)

mostafakamar2308 (approved):

neuodev:
🚀 [Nova](https://nova-bkrf69x1q-litespace-core.vercel.app) 
🚀 [Luna](https://luna-mjplpsdon-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-m4bvpjin5-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-2jmlo8qmb-litespace-core.vercel.app)

neuodev (approved):

neuodev:
🚀 [Nova](https://nova-6lisvjasa-litespace-core.vercel.app) 
🚀 [Luna](https://luna-dklkrisrw-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-j944j7602-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-id7ex09io-litespace-core.vercel.app)

PR #268, opened 2024-12-31
added(sol): availability slots required general logic

mmoehabb:
ClickUp Task: [https://app.clickup.com/t/86978u2tr](https://app.clickup.com/t/86978u2tr)

### DIFF
--- a/packages/sol/src/availabilitySlots.ts
+++ b/packages/sol/src/availabilitySlots.ts
@@ -1,0 +1,190 @@
+import { IAvailabilitySlot } from "@litespace/types";
+import { dayjs } from "@/dayjs";
+import { flatten, orderBy } from "lodash";
+
+/**
+  * Divide list of slots into sub-slots. It is used whenever a user wants to book a lesson
+  * e.g., slot from 8am-10pm, split by 30min => 8-8:30, 8:30-9, 9-9:30, 9:30-10.
+  */
+export function getSubSlotsBatch(
+  /**
+   * list of the slots to be divided.
+   */
+  slots: IAvailabilitySlot.Slot[], 
+  /**
+   * the duration of each subslot in minutes.
+   */
+  duration: number,
+): IAvailabilitySlot.SubSlot[] {
+  return flatten(slots.map(slot => getSubSlots(slot, duration)));
+}
+
+export function getSubSlots(
+  slot: IAvailabilitySlot.Slot,
+  duration: number,
+): IAvailabilitySlot.SubSlot[] {
+  let start = dayjs.utc(slot.start);
+  let end = start.add(duration, "minutes");
+  const subslots: IAvailabilitySlot.SubSlot[] = [];
+  while (end.isBefore(dayjs.utc(slot.end)) || end.isSame(dayjs.utc(slot.end))) {
+    subslots.push({
+      parent: slot.id,
+      start: start.toISOString(),
+      end: end.toISOString(),
+    });
+    start = end;
+    end = start.add(duration, "minutes");
+  }
+  return subslots;
+}
+
+/** 
+  * checks if the passed book info can be booked in the given slot.
+  */
+export function canBook({ 
+  slot,
+  bookedSubslots,
+  bookInfo,
+}: {
+  /**
+   * the slot to be partailly or fully booked.
+   */
+   slot: IAvailabilitySlot.Slot;
+  /**
+   * subslots that are already booked (to be subtracted).
+   */
+   bookedSubslots: IAvailabilitySlot.SubSlot[];
+  /**
+   * the time that's needed to be booked.
+   */
+   bookInfo: {
+    /**
+     * The new slot `start` as iso-utc datetime.
+     */
+     start: string;
+    /**
+     * The new slot duration in minutes.
+     */
+     duration: number;
+   }
+}): boolean {
+  const availableSubSlots = subtractSlots(slot, bookedSubslots);
+  const found = availableSubSlots.find(subslot => isSuperSlot(subslot, {
+    parent: subslot.parent,
+    start: dayjs.utc(bookInfo.start).toISOString(),
+    end: dayjs.utc(bookInfo.start).add(bookInfo.duration, "minutes").toISOString(),
+  }))
+  return !!found;
+}
+
+/**
+ * removes the booked slots (represented by subslots) from the `slots` and 
+ * then returns the free remaining space/time represented by subslots. 
+ */
+export function subtractSlotsBatch({
+  slots,
+  subslots,
+}: {
+   slots: IAvailabilitySlot.Slot[];
+   subslots: IAvailabilitySlot.SubSlot[];
+}): IAvailabilitySlot.SubSlot[] {
+  const res: IAvailabilitySlot.SubSlot[] = [];
+  for (const slot of slots) {
+    const filtered = subslots.filter(subslot => isParent(slot, subslot));
+    res.push(
+      ...subtractSlots(slot, filtered).map(subslot => ({
+        parent: slot.id,
+        start: subslot.start,
+        end: subslot.end
+      }))
+    );
+  }
+  return res;
+}
+
+export function subtractSlots(
+  slot: IAvailabilitySlot.Slot, 
+  subslots: IAvailabilitySlot.SubSlot[]
+): IAvailabilitySlot.SubSlot[] {
+
+  const res: IAvailabilitySlot.SubSlot[] = [];
+  const mainSubSlot: IAvailabilitySlot.SubSlot = {
+    parent: slot.id,
+    start: slot.start,
+    end: slot.end
+  };
+
+  for (const subslot of orderSlots(subslots, "asc")) {
+    const start = dayjs.utc(mainSubSlot.start);
+    const end = dayjs.utc(subslot.start);
+
+    const empty = start.isSame(end);
+    if (!empty) res.push({
+      parent: mainSubSlot.parent, 
+      start: start.toISOString(), 
+      end: end.toISOString(),
+    });
+
+    mainSubSlot.start = subslot.end;
+  }
+
+  const empty = dayjs(mainSubSlot.start).isSame(mainSubSlot.end);
+  if (!empty) res.push(mainSubSlot);
+  return res;
+}
+
+/**
+ * checks if a specific slot intersects with at least one slot of the passed list.
+ * the boundaries are excluded in the intersection.
+ */
+export function isIntersecting(
+  target: IAvailabilitySlot.GeneralSlot, 
+  slots: IAvailabilitySlot.GeneralSlot[]
+): boolean {
+  for (const slot of slots) {
+    // target slot started after the current slot.
+    const startedAfterCurrentSlot = 
+      dayjs.utc(target.start).isAfter(slot.end) || 
+      dayjs.utc(target.start).isSame(slot.end);
+    // target slot eneded before the current slot.
+    const endedBeforeCurrentSlot = 
+      dayjs.utc(target.end).isBefore(slot.start) || 
+      dayjs.utc(target.end).isSame(slot.start);
+    if (!startedAfterCurrentSlot && !endedBeforeCurrentSlot) return true
+  }
+  return false;
+}
+
+/**
+ *  checks if slot `a` is the parent of subslot `b`.
+ */
+export function isParent(
+  a: IAvailabilitySlot.Slot,
+  b: IAvailabilitySlot.SubSlot
+): boolean {
+  return a.id === b.parent && isSuperSlot(a, b);
+}
+
+/**
+ *  checks if slot `a` can be considered as a super slot (contains) for slot `b` or not.
+ */
+export function isSuperSlot(
+  a: IAvailabilitySlot.GeneralSlot, 
+  b: IAvailabilitySlot.GeneralSlot
+): boolean {
+  return !dayjs.utc(a.start).isAfter(b.start) && !dayjs.utc(a.end).isBefore(b.end);
+}
+
+/** 
+ * immutably orders/sorts a list of slots/subslots.
+ */
+export function orderSlots(
+  slots: IAvailabilitySlot.GeneralSlot[], 
+  dir: 'asc' | 'desc'
+): IAvailabilitySlot.GeneralSlot[] {
+  return orderBy(
+    slots, 
+    [ (slot) => dayjs.utc(slot.start) ], 
+    [ dir ]
+  );
+}
--- a/packages/sol/src/index.ts
+++ b/packages/sol/src/index.ts
@@ -15,3 +15,4 @@ export * from "@/user";
 export * from "@/interview";
 export * from "@/session";
 export * from "@/verification";
+export * from "@/availabilitySlots";
--- a/packages/sol/tests/availabilitySlot.test.ts
+++ b/packages/sol/tests/availabilitySlot.test.ts
@@ -1,0 +1,277 @@
+import { IAvailabilitySlot } from "@litespace/types";
+import { dayjs } from "@/dayjs";
+import { 
+  canBook,
+  getSubSlots,
+  getSubSlotsBatch,
+  isIntersecting,
+  isSuperSlot,
+  orderSlots,
+  subtractSlotsBatch
+} from "@/availabilitySlots";
+import { expect } from "chai";
+import { range } from "lodash";
+
+describe("AvailabilitySlot", () => {
+  it("should divide a slot into discrete number of subslots", () => {
+    const now = dayjs.utc();
+    const slot: IAvailabilitySlot.Slot = {
+      id: 1,
+      start: now.toISOString(),
+      end: now.add(2, "hours").toISOString(),
+    }
+
+    let subslots = getSubSlots(slot, 15);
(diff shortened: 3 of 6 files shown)


neuodev:
🚀 [Nova](https://nova-80iyp56t6-litespace-core.vercel.app) 
🚀 [Luna](https://luna-7ze162hle-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-geyqkulnf-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-d6wfaxgg4-litespace-core.vercel.app)

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,144 @@
...
+import { dayjs } from "@/dayjs";
+import { clone, flatten, min } from "lodash";
+
+/**
+  * Divide list of slots into sub-slots. It is used whenever a user wants to book a lesson
+  * e.g., slot from 8am-10pm, split by 30min => 8-8:30, 8:30-9, 9-9:30, 9:30-10.
+  * @param slots - list of the slots to be divided.
+  * @param duration - the duration of each subslot in minutes.
+  */
+export function getSubslotsBatch(
```
neuodev, line 11: Let's stick to `SubSlots` instead of `Subslots` for readability.  In the description we can write `sub-slot` instead of `subslot`. 

```suggestion
export function getSubSlotsBatch(
```

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,144 @@
...
+  slots: IAvailabilitySlot.Slot[],
```
neuodev, line 12: I believe it is better to move the params description to be above the param itself.  The same applies to all other params.

```suggestion
  /**
   * List of the slots to be divided
   */
  slots: IAvailabilitySlot.Slot[], 
```

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,144 @@
...
+  const bookingSlot: IAvailabilitySlot.SubSlot = {
+    parent: -1,
+    start: bookInfo.start,
+    end: dayjs(bookInfo.start).add(bookInfo.duration, "minutes").toISOString(),
+  };
+  for (const slot of filtered) {
+    if (isSuperSlot(slot, bookingSlot))
+      return true;
+  }
+  return false;
```
neuodev, line 69: This is not how you check if you the new slot will fit in the remaining space in the parent slot. 

Here is the steps to do this. 
1. Subtract the parent slots from all subslots. The result will be all free sub-slots. 
2. Check if the new slot (`bookInfo`) fit in one or more sub-slot. It if does fit, this mean you can book this new slot.

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,144 @@
...
+  const cloned = clone(slots);
+  return cloned.sort((a, b) => {
+    const dirValue = dir === "asc" ? 1 : -1;
+    if (dayjs.utc(a.start).isBefore(b.start))
+      return -1 * dirValue;
+    else if (dayjs.utc(a.start).isAfter(b.start))
+      return 1 * dirValue
+    else
+      return 0;
+  });
```
neuodev, line 143: The function is hard to understand from the first look. Let's use lodash `orderBy` for convenience. It is used [here](https://github.com/litespace-org/litespace/blob/ddc09149172b24c29ab774f769854ede9c66860a/services/server/src/lib/tutor.
ts#L105).

neuodev (changes requested):

neuodev:
🚀 [Nova](https://nova-exweke196-litespace-core.vercel.app) 
🚀 [Luna](https://luna-1qyjkaby4-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-kbaf6urij-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-d1aa9v5se-litespace-core.vercel.app)

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,148 @@
...
+  return subslots;
+}
+
+/** 
+  * checks if the passed book info can be booked in the given slot.
+  */
+export function canBook({ 
+  /**
+   * the slot to be partailly or fully booked.
+   */
```
neuodev, line 49: The JS Docs should be above the types.

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,148 @@
...
+  bookedSubslots,
+  /**
+   * the time that's needed to be booked.
+   */
+  bookInfo,
+}: {
+   slot: IAvailabilitySlot.Slot;
+   bookedSubslots: IAvailabilitySlot.SubSlot[];
+   bookInfo: {
+     start: string;
```
neuodev, line 63:
```suggestion
    /**
    * The new slot start as iso-utc datetime.
    */
     start: string;
```

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,148 @@
...
+     duration: number; // in minutes
```
neuodev, line 64:
```suggestion
    /**
    * The new slot duration in minutes.
    */
     duration: number;
```

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,148 @@
...
+  subslots,
+}: {
+   slots: IAvailabilitySlot.Slot[];
+   subslots: IAvailabilitySlot.SubSlot[];
+}): IAvailabilitySlot.SubSlot[] {
+  const res: IAvailabilitySlot.SubSlot[] = [];
+  for (const slot of slots) {
+    const filtered = subslots.filter(subslot => isSuperSlot(slot, subslot));
+    res.push(
+      ...Schedule.splitBy(slot, filtered).map(event => ({
```
neuodev, line 94: `Schedule` belongs to the old logic (rules) and it will be removed once we move to slots. Please write a similar logic here using our new terminology.

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,148 @@
...
+        start: event.start,
+        end: event.end
+      }))
+    );
+  }
+  return res;
+}
+
+// checks if a specific slot intersects with at least one slot of the passed list.
+// the boundaries are included in the intersection.
```
neuodev, line 105: JS Docs should be multi-line comment.

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,148 @@
...
+export function isIntersecting(
+  target: IAvailabilitySlot.GeneralSlot, 
+  slots: IAvailabilitySlot.GeneralSlot[]
+): boolean {
```
neuodev, line 109: We can improve the function signature by applying these changes.

```suggestion
export function isIntersecting<T extends {start: string, end: string}>(
  target: T, 
  slots: T[]
): boolean {
```

This way the function is generic and we are not limited by the `GeneralSlot` type. 

> [!NOTE]
> Don't forget to remove `GeneralSlot` in case we don't need it.

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (startTimeCondition) continue;
+
+    const endTimeCondition = dayjs.utc(target.end).isBefore(slot.start) || 
+      dayjs.utc(target.end).isSame(slot.start);
+    if (endTimeCondition) continue;
+
+    return true;
+  }
+  return false;
+}
```
neuodev, line 122: We can use more sophisticated logic like [this one](https://github.com/litespace-org/litespace/blob/aebee0e646360ee861098e57b12dd029098fca1a/packages/sol/src/rule.ts#L216-L222).

Here is a stackoverflow question about who to solve the [overlap problem](https://stackoverflow.com/questions/3269434/whats-the-most-efficient-way-to-test-if-two-ranges-overlap). 

image

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ *  checks if slot `a` can be considered as a super slot (contains) for slot `b` or not.
+ */
+export function isSuperSlot(
+  a: IAvailabilitySlot.GeneralSlot, 
+  b: IAvailabilitySlot.GeneralSlot
```
neuodev, line 129:
```suggestion
/**
 *  checks if slot `a` can be considered as a super slot (contains) for slot `b` or not.
 */
export function isSuperSlot(
  a: IAvailabilitySlot.Slot, // `SubSlot` can never be a parent to another subslot!  
  b: IAvailabilitySlot.SubSlot
```

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,148 @@
...
+): boolean {
+  if (dayjs.utc(a.start).isAfter(b.start)) 
+    return false;
+  if (dayjs.utc(a.end).isBefore(b.end)) 
+    return false;
+  return true;
```
neuodev, line 135: * You can make the code more clean (give it a try)
* You also need to verify the `parentId`. `slot.id === subslot.parentId`.

neuodev (changes requested):

packages/sol/src/availabilitySlots.ts
mmoehabb: Using generic here is not quite feasible; if the first parameter type is `Slot` then the second has to be an array of `Slots`, and this is not the case in some use cases (wherein the second parameter is a list of `SubSlot`s). This has an easy workaround, but it'll unnecessarily make the code ugly. The code was initially like that: `T extends IAvailabilitySlot.Slot | IAvailbilitySlot.SubStot`, but then I considered cleaning this generic verbose ugliness!

packages/sol/src/availabilitySlots.ts
mmoehabb: `isSuperSlot` doesn't look in the parent-child relation. It checks if some slot/subslot contains another. Just like in set theory; a set is a subset of itself, and a subset can be a superset of subsets.

It's being in handy (with this signature) in the `canBook` function.

packages/sol/src/availabilitySlots.ts
mmoehabb: For the second check, I propose a function to create slots and subslots. It can easily be misused or forged.

packages/sol/src/availabilitySlots.ts
mmoehabb: The logic in the screenshot doesn't meet the requirement of our use cases. We shall exclude the boundaries; meaning if the end of _x_ is the same as the start of _y_, we still consider them NOT intersecting.

> I just noticed that I wrote in the comment "The boundaries are includes", that's a mistake, they are excluded from the intersection.

I'm not sure what do you mean by the word "sophisticated". But the code [https://github.com/litespace-org/litespace/blob/aebee0e646360ee861098e57b12dd029098fca1a/packages/sol/src/rule.ts#L216-L222](here) is just like my first attempt to solve the problem, before trying to make it more efficient (using `continue`).

I've revised my code again and enhanced it for the sake of readability. Give it another look.

neuodev:
🚀 [Nova](https://nova-3yyifu8ak-litespace-core.vercel.app) 
🚀 [Luna](https://luna-lxfky2tx4-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-ghffaikuw-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-bp3djlyoe-litespace-core.vercel.app)

packages/sol/src/availabilitySlots.ts
mmoehabb: Or we can remove the id. What is it needed for?... I mean If a subslot _b_ is inside a slot _a_ but has a parent id that doesn't match the id of _a_, how could that matter or make a difference in any sense?

neuodev:
🚀 [Nova](https://nova-3yyifu8ak-litespace-core.vercel.app) 
🚀 [Luna](https://luna-lxfky2tx4-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-pnvnzp73b-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-i8dtig6nb-litespace-core.vercel.app)

packages/sol/src/availabilitySlots.ts
neuodev: The `subslot` in this case might not belong to the parent slot or it is for another user.

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,188 @@
...
+    const filtered = subslots.filter(subslot => isSuperSlot(slot, subslot));
+    res.push(
+      ...subtractSlots(slot, filtered).map(subslot => ({
+        parent: slot.id,
+        start: subslot.start,
+        end: subslot.end
+      }))
+    );
+  }
+  console.log(res)
```
neuodev, line 102: Left over.

packages/sol/src/availabilitySlots.ts
neuodev: Aha, got it.

packages/sol/src/availabilitySlots.ts
mmoehabb: Aha got it. Each tutor has his own slots, that what I was missing.

neuodev:
🚀 [Nova](https://nova-hej53265g-litespace-core.vercel.app) 
🚀 [Luna](https://luna-1gzpxboey-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-hzhe2p3x2-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-al6t9vmce-litespace-core.vercel.app)

packages/sol/src/availabilitySlots.ts
neuodev: Aha, got it. I will take a look one more time into the implementation.

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,197 @@
...
+    // then there is no intersecting and continue (skip) to the next slot.
+    notIntersecting = 
+      dayjs.utc(target.end).isBefore(slot.start) || 
+      dayjs.utc(target.end).isSame(slot.start);
+    if (notIntersecting) continue;
+
+    // Otherwise return true indicating intersecting
+    return true;
+  }
+  return false;
```
neuodev, line 162: * No need to re-assign variables.
* We can make the code more cleaner. 

```suggestion
  for (const slot of slots) {
    // target slots started after the current slot.
    const startedAfterCurrentSlot = 
      dayjs.utc(target.start).isAfter(slot.end) || 
      dayjs.utc(target.start).isSame(slot.end);

    // target slots eneded before the current slot.
    const endedBeforeCurrentSlot = 
      dayjs.utc(target.end).isBefore(slot.start) || 
      dayjs.utc(target.end).isSame(slot.start);

    if (startedAfterCurrentSlot || endedBeforeCurrentSlot) return true
  }

  return false;
```

packages/sol/src/availabilitySlots.ts
```diff
@@ -0,0 +1,197 @@
...
+  return res;
+}
+
+export function subtractSlots(
+  slot: IAvailabilitySlot.Slot, 
+  subslots: IAvailabilitySlot.SubSlot[]
+): IAvailabilitySlot.SubSlot[] {
+
+  const res: IAvailabilitySlot.SubSlot[] = [];
+  let mainSubSlot: IAvailabilitySlot.SubSlot = {
```
neuodev, line 111:
```suggestion
  const mainSubSlot: IAvailabilitySlot.SubSlot = {
```

neuodev (commented):
Greate work! I left a minor comment. We can merge this branch once it is addressed.

neuodev:
@mmoehabb please rebase.

neuodev:
🚀 [Nova](https://nova-k1ps2qc3d-litespace-core.vercel.app) 
🚀 [Luna](https://luna-o5cn1h8gq-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-ly980hgyr-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-glvg9d7e1-litespace-core.vercel.app)

packages/sol/src/availabilitySlots.ts
neuodev: Ah, I spotted the issue. The logic above is about "not intersecting". It should return `false` then.

neuodev:
```ts
    if (!startedAfterCurrentSlot && !endedBeforeCurrentSlot) return true
```

neuodev (approved):

packages/sol/src/availabilitySlots.ts
mmoehabb: Aha got it.

neuodev:
🚀 [Nova](https://nova-jeodg2aod-litespace-core.vercel.app) 
🚀 [Luna](https://luna-4lwc31v39-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-3j33k60nl-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-mz7dklduo-litespace-core.vercel.app)

PR #270, opened 2025-01-01
modif(model): retrieving messages order has been reversed.

mmoehabb:
ClickUp Task: [https://app.clickup.com/t/86978y20k](https://app.clickup.com/t/86978y20k)

### DIFF
--- a/packages/models/src/messages.ts
+++ b/packages/models/src/messages.ts
@@ -141,8 +141,8 @@ export class Messages {
       .clone()
       .select()
       .orderBy([
-        { column: this.column("created_at"), order: "desc" },
-        { column: this.column("id"), order: "desc" },
+        { column: this.column("created_at"), order: "asc" },
+        { column: this.column("id"), order: "asc" },
       ]);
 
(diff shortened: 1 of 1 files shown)


apps/nova/src/lib/room.test.ts
```diff
@@ -9,6 +9,7 @@ describe(nameof(asOtherMember), () => {
     id: 1,
     email: faker.internet.email(),
     name: faker.person.fullName(),
+    bio: "",
```
neuodev, line 12: Let's add some value to verify that is working as expected. Also add for the `other` member.

packages/types/src/room.ts
```diff
@@ -35,6 +35,7 @@ export type PopulatedMemberRow = {
   muted: MemberRow["muted"];
   email: IUser.Row["email"];
   name: IUser.Row["name"];
+  bio: ITutor.Row["bio"];
```
neuodev, line 38: Can figure out the issue at this line?

PR #271, opened 2025-01-01
feat: update spinner & error component

neuodev:
[Task](https://app.clickup.com/t/86977a9h4)


neuodev:
This pull request should be a replacement for #254

neuodev:
🚀 [Nova](https://nova-4810bm6qf-litespace-core.vercel.app) 
🚀 [Luna](https://luna-3q9f372f6-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-kq66c0j3t-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-db2tmltlx-litespace-core.vercel.app)

apps/nova/src/components/UpcomingLessons/Content.tsx
```diff
@@ -26,7 +26,8 @@ export const Content: React.FC<{
   error: boolean;
   hasMore: boolean;
   more: Void;
-}> = ({ list, loading, fetching, error, hasMore, more }) => {
+  refetch: Void;
```
mmoehabb, line 29: The type `Void` bestows ambiguity into the code. It'll be tedious for newcomer teammates. At least we can add a comment describing the expected values to be passed.

apps/nova/src/components/UpcomingLessons/Content.tsx
```diff
@@ -60,10 +61,23 @@ export const Content: React.FC<{
     []
   );
 
-  if (loading) return <Loading className="h-[30vh]" />;
+  if (loading)
+    return (
+      <div className="mt-[133px]">
+        <Loader size="lg" text={intl("upcoming-lessons.loading")} />;
```
mmoehabb, line 67: Remove the semi-colon in the end of the line.

apps/nova/src/components/UpcomingLessons/Content.tsx
```diff
@@ -60,10 +61,23 @@ export const Content: React.FC<{
     []
   );
 
-  if (loading) return <Loading className="h-[30vh]" />;
+  if (loading)
+    return (
+      <div className="mt-[133px]">
```
mmoehabb, line 66: I think using absolute values for margin here is not good design choice. We should use at least "em" instead of "px". I propose using `15vh`.

packages/assets/assets/spinner2.svg
mmoehabb, line 1: Great! I love its simplicity.

mmoehabb (changes requested):

packages/assets/assets/spinner2.svg
mostafakamar2308: Nice work. it works as expected

packages/luna/src/components/Loading/Loader.tsx
```diff
@@ -1,49 +1,21 @@
...
 import { Typography } from "@litespace/luna/Typography";
 import cn from "classnames";
-import { motion } from "framer-motion";
+import Spinner from "@litespace/assets/Spinner2";
 
 export const Loader: React.FC<{
   text?: string;
-  variant?: "small" | "large";
-}> = ({ text, variant = "large" }) => {
+  size?: "sm" | "md" | "lg";
```
mostafakamar2308, line 8: using shorthand like "sm" or "lg" isn't the standard, we usually use "small", "large" in variants. WDYT?

packages/assets/package.json
```diff
@@ -102,6 +102,7 @@
     "./SoundSmall": "./dist/SoundSmall.tsx",
     "./Sound": "./dist/Sound.tsx",
     "./Spinner": "./dist/Spinner.tsx",
+    "./Spinner2": "./dist/Spinner2.tsx",
```
mostafakamar2308, line 105: I think we will replace the spinner in all the app, we can discuss this with zyad.
if so, then we can have only 1 spinner

packages/luna/src/components/Loading/Loader.stories.tsx
```diff
@@ -23,23 +23,30 @@ const meta: Meta<Component> = {
...
   },
 };
-export const SmallLoader: StoryObj<Component> = {
+
+export const WithText: StoryObj<Component> = {
   args: {
     text: faker.lorem.words(10),
-    variant: "small",
   },
 };
```
mostafakamar2308, line 51: Can you add another story with no text?

packages/luna/src/components/Loading/LoadingError.tsx
```diff
@@ -4,15 +4,26 @@ import { useFormatMessage } from "@/hooks/intl";
...
+          {
+            "tw-p-[3.33px] tw-w-10 tw-h-10": size === "sm",
+            "tw-p-[5.33px] tw-w-16 tw-h-16": size === "md",
+            "tw-p-[6.67px] tw-w-20 tw-h-20": size === "lg",
+          }
+        )}
+      >
         <ExclaimationMarkCircle />
       </div>
       <Typography
```
mostafakamar2308, line 29: in the small variant of this component the typography buttons will have some minor differences.

Small Variant:
image

Large Variant:
image

We can ask zyad about the medium variant typography and button size.

mostafakamar2308 (changes requested):
Really great and clean work ❤️ 
I left small comments, please resolve them.

apps/nova/src/components/UpcomingLessons/Content.tsx
neuodev: The type `Void` is exactly the same as `() => void`. It doesn't accept anything. **The main idea behind it is represent a simple callback handler that doesn't accept any params**. 

I think the problem is with the name. It conflicts with the native type `void` (lower case). Maybe we can rename the type into something else like `Callback` or not use it at all and just type `() => void`. 

Or do it this way (the best in my opinion). 

```suggestion
  refetch(): void;
```

cc @mmoehabb @mostafakamar2308

apps/nova/src/components/UpcomingLessons/Content.tsx
neuodev: I would like to have a discussion about where should the spinner should be located. I believe we don't need to render the spinner at the middle of the page. We want to render it at the top 1/3 of the screen (or its parent). In this case we can make the margin in percentage `mt-[calc(100%-spinner_height)]` . WDYT? @mmoehabb @mostafakamar2308

packages/luna/src/components/Loading/Loader.tsx
neuodev: I got you! Good! I will revert it back.

packages/assets/package.json
neuodev: Done.

packages/luna/src/components/Loading/Loader.tsx
neuodev: Done.

packages/assets/package.json
neuodev: Great note! I noticed that the first spinner (the old one) is used in the book lesson dialog. Ziad replaced the old spinner with [the new one](https://www.figma.com/design/4SwhwulDQ3eEDqevWprW9E/LiteSpace?node-id=4049-22326&t=sGBbfnYz8BtZjDQa-4). 

I will remove the old spinner and use the new one according to the design.

neuodev: Done.

packages/luna/src/components/Loading/Loader.stories.tsx
neuodev: Done. I added all there sizes with text.

packages/luna/src/components/Loading/LoadingError.tsx
neuodev: I think Ziad should add it to the design system to make it clear. I will create a task for him and we can implement these changes in a follow up. [Task](https://app.clickup.com/t/86979jxb2)

apps/nova/src/components/UpcomingLessons/Content.tsx
neuodev: I did some experiments and I found that setting a margin top of `15vh` is the best option.

neuodev: It will adapt to the user view/screen.

https://github.com/user-attachments/assets/764a1ccf-f4c5-43bb-a30f-40b24f3ba072

neuodev:
🚀 [Nova](https://nova-45ikwr7jh-litespace-core.vercel.app) 
🚀 [Luna](https://luna-mw3myhyja-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-61ty745bf-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-3q12moifc-litespace-core.vercel.app)

apps/nova/src/components/UpcomingLessons/Content.tsx
mmoehabb: You didn't get what I mean. Both `Void` and `() => void` bestow ambiguity. When I first read this code I didn't easily figure out what value should I pass the parameter `retry`. We may define more specific type or add a simple comment, for instance: `// A callback function that invokes tanstack query refetch method`.

mmoehabb: Maybe the best solution is to define a "dump" type in the types dir:
```TS
/**
 * A callback function that invokes tanstack query refetch method
 */
type TanStackRefetchCallback = Void;
```
what do you think? @neuodev

apps/nova/src/components/UpcomingLessons/Content.tsx
```diff
@@ -60,10 +61,23 @@ export const Content: React.FC<{
...
+      <div className="mt-[15vh]">
+        <Loader size="large" text={intl("upcoming-lessons.loading")} />
+      </div>
+    );
 
-  // todo: add error element
-  if (error) return "TODO: ERROR";
+  if (error)
+    return (
+      <div className="mt-[133px]">
```
mmoehabb, line 73: It should have the same margin as the loader: `mt-[15vh]`.

mmoehabb (changes requested):

apps/nova/src/components/UpcomingLessons/Content.tsx
neuodev: > The child component should not be considered with the parent implementation 

In this case, the `Content` component should not be aware of how the parent is fetching the data. For example, the parent should use what ever approach possible (e.g., tanstack query, Fetch API, axios, ...etc).

neuodev:
🚀 [Nova](https://nova-lv0nt6vsp-litespace-core.vercel.app) 
🚀 [Luna](https://luna-k4kmzm9e0-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-iy2uqtg9n-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-5gk22x0d5-litespace-core.vercel.app)

mmoehabb (approved):

neuodev:
🚀 [Nova](https://nova-ckly88yk7-litespace-core.vercel.app) 
🚀 [Luna](https://luna-g7qfyb5nh-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-4o2kt8tmo-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-91w7ouc63-litespace-core.vercel.app)

mostafakamar2308 (approved):

PR #272, opened 2025-01-02
feat: pre-join session wss event has been added. And session handler modified to utilize the acknowlegment mechanism.

mmoehabb:
ClickUp Task: [https://app.clickup.com/t/86979r2et](https://app.clickup.com/t/86979r2et)

### DIFF
--- a/services/server/src/lib/session.ts
+++ b/services/server/src/lib/session.ts
@@ -1,11 +1,10 @@
 import { unpackRules } from "@litespace/sol/rule";
-import { IInterview, ILesson, IRule } from "@litespace/types";
+import { IInterview, ILesson, IRule, Wss } from "@litespace/types";
 import { concat, first, isEmpty } from "lodash";
 import dayjs from "@/lib/dayjs";
 import { platformConfig } from "@/constants";
 import { interviews, lessons } from "@litespace/models";
-import { INTERVIEW_DURATION } from "@litespace/sol";
-import { asSessionId } from "@litespace/sol";
+import { asSessionId, getSessionType } from "@litespace/sol";
 
 // todo: impl: each tutor can have interview each 3 months.
 export function canBeInterviewed(sessions: IInterview.Self[]): boolean {
@@ -64,46 +63,45 @@ export function canBook({
 }
 
 // todo: write tests
-export async function canJoinSession({
+/**
+ * this function has been written to clean up the wss session handler
+ * by seperating this verbose details from the main logic code.
+ */
+export async function canJoinSessionAck({
   userId,
   sessionId,
 }: {
   userId: number;
   sessionId: string;
-}) {
-  const now = dayjs.utc();
-  const sessionType = first(sessionId.split(":"));
+}): Promise<Wss.AcknowledgePayload> {
+  const sessionType = getSessionType(sessionId);
 
   if (sessionType === "lesson") {
     const lesson = await lessons.findBySessionId(asSessionId(sessionId));
-    if (!lesson) return false;
-
-    const start = dayjs.utc(lesson.start);
-    const end = start.add(lesson.duration, "minutes");
-    // todo: unmagic "10" minutes
-    // const early = start.isAfter(now) && start.diff(now, "minutes") > 10;
-    // if (end.isAfter(now) || early) return false;
+    if (!lesson) return { code: Wss.AcknowledgeCode.LessonNotFound };
 
     const members = await lessons.findLessonMembers([lesson.id]);
     const member = members.find((member) => member.userId === userId);
-    if (!member) return false;
+    if (!member)
+      return {
+        code: Wss.AcknowledgeCode.Unallowed,
+        message: "The user is not a member of the lesson.",
+      };
 
-    return true;
+    return { code: Wss.AcknowledgeCode.Ok };
   }
 
   const interview = await interviews.findBySessionId(asSessionId(sessionId));
-  if (!interview) return false;
-
-  const start = dayjs.utc(interview.start);
-  const end = start.add(INTERVIEW_DURATION, "minutes");
-  // todo: unmagic "10" minutes
-  const early = start.isAfter(now) && start.diff(now, "minutes") > 10;
-  if (end.isAfter(now) || early) return false;
+  if (!interview) return { code: Wss.AcknowledgeCode.InterviewNotFound };
 
   const member =
     interview.ids.interviewer === userId ||
     interview.ids.interviewee === userId;
-  if (!member) return false;
+  if (!member)
+    return {
+      code: Wss.AcknowledgeCode.Unallowed,
+      message: "The user is not a member of the lesson.",
+    };
 
-  return true;
+  return { code: Wss.AcknowledgeCode.Ok };
 }
--- a/services/server/src/wss/handlers/connection.ts
+++ b/services/server/src/wss/handlers/connection.ts
@@ -1,8 +1,8 @@
 import { isAdmin, isGhost, isStudent, isTutor } from "@litespace/auth";
-import { dayjs, logger, safe } from "@litespace/sol";
+import { logger, safe } from "@litespace/sol";
 import { Wss } from "@litespace/types";
 import { WssHandler } from "@/wss/handlers/base";
-import { interviews, lessons, rooms } from "@litespace/models";
+import { rooms } from "@litespace/models";
 import { background } from "@/workers";
 import { PartentPortMessage, PartentPortMessageType } from "@/workers/messages";
 import { asSessionRoomId, asChatRoomId } from "@/wss/utils";
@@ -25,9 +25,14 @@ export class Connection extends WssHandler {
 
       await cache.onlineStatus.addUser(user.id);
       this.announceStatus({ userId: user.id, online: true });
+      this.joinChatRooms();
 
-      await this.joinRooms();
-      if (isAdmin(this.user)) this.emitServerStats();
+      if (isStudent(this.user)) {
+        this.socket.join(Wss.Room.TutorsCache);
+      } else if (isAdmin(this.user)) {
+        this.socket.join(Wss.Room.ServerStats);
+        this.emitServerStats();
+      }
     });
     if (error instanceof Error) stdout.error(error.message);
   }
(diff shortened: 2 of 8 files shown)


neuodev:
🚀 [Nova](https://nova-pnex1la8v-litespace-core.vercel.app) 
🚀 [Luna](https://luna-p6zd4cg11-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-l6e0ynq4x-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-5no5elbli-litespace-core.vercel.app)

services/server/src/wss/handlers/session.ts
```diff
@@ -11,83 +11,116 @@ import { sessionId } from "@/validation/utils";
...
     this.socket.on(Wss.ClientEvent.JoinSession, this.onJoinSession.bind(this));
     this.socket.on(Wss.ClientEvent.LeaveSession, this.onLeaveSession.bind(this));
     return this;
   }
 
+  /*
+   *  This event listener shall be called whenever a user
+   *  renders the pre-join page in the frontend.
+   *  It joins the user socket to the session socket room.
+   */
```
neuodev, line 28:
```suggestion
  /*
   *  This event listener shall be called whenever a user
   *  wants to get updates about the current session (e.g., current members)  
   *  but without joining the session. It adds the user socket to the session socket room.
   */
```

services/server/src/wss/handlers/session.ts
```diff
@@ -11,83 +11,116 @@ import { sessionId } from "@/validation/utils";
...
-      // todo: add ghost as a member of the session
-      if (!user) return stdout.error("user undefined!");
-      if (isGhost(user)) return stdout.warning("Unsupported");
+      const ack = await this.canJoinSession(sessionId);
+      if (ack.code !== Wss.AcknowledgeCode.Ok)
+        return this.call(callback, ack);
 
-      stdout.info(`User ${user.id} is joining session ${sessionId}.`);
+      // ensured to be self from the canJoinSession function
+      const user = this.user as IUser.Self;
```
neuodev, line 58: Let's avoid casting and perform the check one more time. 

```suggestion
      // ensured to be self from the canJoinSession function
      const user = this.user
      if (!isUser(user)) return 
```

services/server/src/lib/session.ts
```diff
@@ -64,46 +64,68 @@ export function canBook({
...
-    // const early = start.isAfter(now) && start.diff(now, "minutes") > 10;
-    // if (end.isAfter(now) || early) return false;
+    if (end.isBefore(now)) return {
+      code: Wss.AcknowledgeCode.Unallowed,
+      message: "The lesson is closed."
+    };
+    if (start.isAfter(now) && start.diff(now, "minutes") > 10) return {
+      code: Wss.AcknowledgeCode.Unallowed,
+      message: "The lesson is not about to start yet."
+    };
```
neuodev, line 95: Based on the new business decisions, user can join the session any time (before, during or after). We will just show him a message an appropriate message.

neuodev: Just verify the user can be a member of the lesson/interview.

services/server/src/lib/session.ts
```diff
@@ -64,46 +64,68 @@ export function canBook({
...
+export async function canJoinSessionAck({
   userId,
   sessionId,
 }: {
   userId: number;
   sessionId: string;
-}) {
+}): Promise<Wss.AcknowledgePayload> {
   const now = dayjs.utc();
   const sessionType = first(sessionId.split(":"));
```
neuodev, line 79: It is better to move this function to `@litespace/sol`. 

```ts
function getSessionType(sessionId: ISession.Id): ISession.Type | null;
```
```suggestion
  const sessionType = getSessionType(sessionId)
```

services/server/src/wss/handlers/connection.ts
```diff
@@ -25,9 +25,15 @@ export class Connection extends WssHandler {
...
-      if (isAdmin(this.user)) this.emitServerStats();
+      this.joinChatRooms();
+
+      if (isStudent(this.user)) {
+        this.socket.join(Wss.Room.TutorsCache);
+      }
+      else if (isAdmin(this.user)) {
+        this.socket.join(Wss.Room.ServerStats);
+        this.emitServerStats();
+      }
```
neuodev, line 36: It is better to use list of `if` statements instead of `if/else`. 

```suggestion
      if (isStudent(this.user)) {
        this.socket.join(Wss.Room.TutorsCache);
      }
      
      if (isAdmin(this.user)) {
        this.socket.join(Wss.Room.ServerStats);
        this.emitServerStats();
      }
```

services/server/src/wss/handlers/session.ts
```diff
@@ -11,83 +11,116 @@ import { sessionId } from "@/validation/utils";
...
-        sessionId,
+        sessionId: asSessionId(sessionId)
       });
 
-      if (!canJoin) throw Error("Forbidden");
-
-      // add user to the session by adding its id in the cache
-      await cache.session.addMember({ userId: user.id, sessionId: asSessionId(sessionId) });
+      // in case the user somehow has joined the room before
+      // the pre-join step.
```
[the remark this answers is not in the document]
mmoehabb: We should take in account malicious users, bugs or bad code in the frontend. It  should never happen, but it could happen. We can either do this workaround or prevent the user from doing so. The former (which I chose)  is straightforward, the latter, on the other hand, requires us to store in the cache the uses who are waiting in the pre-join (which I found unnecessary). What do you think?

services/server/src/wss/handlers/connection.ts
mmoehabb: `if/else` is more efficient. Can the admin be a student?
I know this is just _slightly_ efficient... but I am putting in my mind that thousands of users maybe using this function daily so I'm doing the best I can.

services/server/src/lib/session.ts
mmoehabb: Hmm... after the session is kinda exotic, so to speak. Why would the user join after the session?

neuodev:
🚀 [Nova](https://nova-75w4y21en-litespace-core.vercel.app) 
🚀 [Luna](https://luna-gsvaoze6r-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-5jpwhxg9h-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-e8hdrdm6w-litespace-core.vercel.app)

neuodev:
🚀 [Nova](https://nova-75w4y21en-litespace-core.vercel.app) 
🚀 [Luna](https://luna-gsvaoze6r-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-hl3fykiv7-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-2nqdva31b-litespace-core.vercel.app)

services/server/src/wss/handlers/connection.ts
neuodev: > Can the admin be a student?

It will never happen.

services/server/src/lib/session.ts
neuodev: He might click the link accidentally or opened the wrong link.

services/server/src/wss/handlers/session.ts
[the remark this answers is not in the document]
neuodev: If a client wants to interacts with the server, it must be compliant with how it works. The server **defines** how the client show work via its API (e.g., socket events, RESTful API).

> the latter, on the other hand, requires us to store in the cache the users who are waiting in the pre-join 

No need to store them. At the moment we don't need to do this.

services/server/src/lib/session.ts
neuodev: We don't want to be strict or restrictive. We want to be fault tolerant.

services/server/src/wss/handlers/session.ts
```diff
@@ -11,83 +11,116 @@ import { sessionId } from "@/validation/utils";
...
+  async onPreJoinSession(data: unknown, callback?: Wss.AcknowledgeCallback) {
+    const result = await safe(async () => {
+      const { sessionId } = generalSessionPayload.parse(data);
+
+      const ack = await this.canJoinSession(sessionId);
+      if (ack.code !== Wss.AcknowledgeCode.Ok)
+        return this.call(callback, ack);
+
+      // join the user socket to the corresponding session socket.io room
+      this.socket.join(asSessionRoomId(sessionId));
```
neuodev, line 38: Let's send an `ack` back that the event was okay.

services/server/src/wss/handlers/connection.ts
mmoehabb: Then we may use `if/else if`.

services/server/src/lib/session.ts
mmoehabb: I think showing a message to the user saying that a session has already been closed is "fault tolerant" in this case, isn't it?
What concerns me is that users can use old sessions to communicate, which I guess should not be allowed by us (the server).

neuodev:
🚀 [Nova](https://nova-278gwrxou-litespace-core.vercel.app) 
🚀 [Luna](https://luna-dpr75oqj1-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-cnr788san-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-oc7ygapqx-litespace-core.vercel.app)

neuodev:
🚀 [Nova](https://nova-278gwrxou-litespace-core.vercel.app) 
🚀 [Luna](https://luna-dpr75oqj1-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-cnr788san-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-fc51v9h0m-litespace-core.vercel.app)

services/server/src/lib/session.ts
neuodev: Let's just enable it for now (for testing) and we can come back to it later.

neuodev (approved):

neuodev:
🚀 [Nova](https://nova-66pmnd0yt-litespace-core.vercel.app) 
🚀 [Luna](https://luna-5rhgmpjjw-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-hfnt19tx9-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-hdznq5gwn-litespace-core.vercel.app)

PR #304, opened 2025-01-17
feat(nova): implemented tutor profile settings

mostafakamar2308:


neuodev:
🚀 [Nova](https://nova-kwj3tx5gg-litespace-core.vercel.app) 
🚀 [Luna](https://luna-l1opc6dmm-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-hgops57vm-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-f7fp32ifr-litespace-core.vercel.app)

mmoehabb (changes requested):

packages/luna/src/locales/ar-eg.json
```diff
@@ -477,6 +477,12 @@
   "chat.message.error": " عذرًا، حدث خطأ أثناء تحميل رسائل المحادثة، برجاء المحاولة مرة أخرى",
   "chat.error": " عذرًا، حدث خطأ أثناء تحميل المحادثات، برجاء المحاولة مرة أخرى",
   "tutor-settings.profile.title": "الملف الشخصى",
+  "tutor-settings.personal-info.title": "المعلومات الشخصية",
+  "tutor-settings.personal-info.name": "الاسم بالكامل",
+  "tutor-settings.personal-info.bio": "النبذة المختصرة",
+  "tutor-settings.personal-info.topics": "مجالات خبرة المعلم",
+  "tutor-settings.personal-info.about": "السيرة الذاتية",
+  "tutor-settings.personal-info.video": "الفيديو التعريفى",
```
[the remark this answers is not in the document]
neuodev: it should be part of the `tutor-settings`. The tutor profile is the "public" profile information which is viewed by students.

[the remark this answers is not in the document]
mostafakamar2308: `tutor.profile` is for tutor profile page (where a student sees them)

apps/nova/src/components/Common/Animate.tsx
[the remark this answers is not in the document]
mostafakamar2308: this warning
image

the main problem is that key is a reserved prop name for react, so I changed it

apps/nova/src/components/Common/Animate.tsx
```diff
@@ -1,13 +1,13 @@
 import React from "react";
```
neuodev, line 1: There is a component called [`Animate`](https://github.com/litespace-org/litespace/blob/master/packages/luna/src/components/Animate/Animate.tsx) which used to animate the opacity and the height of the children. You can find how it is being used in the `TopicSelectionDialog`.


You can create component and call it `AnimateOpacity` in `luna/src/components/Animate/AnimateOpacity.tsx` and re-use it here.

apps/nova/src/components/Common/Animate.tsx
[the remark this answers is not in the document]
neuodev: No need for `name` prop. You can just use the `key`. No need to access the key be the `Animate` component.

apps/nova/src/components/Common/Animate.tsx
```diff
@@ -1,13 +1,13 @@
...
-export const Animate: React.FC<{ children: React.ReactNode; key: string }> = ({
+export const Animate: React.FC<{ children: React.ReactNode; name: string }> = ({
   children,
-  key,
+  name,
 }) => {
   return (
     <motion.div
-      key={key}
+      key={name}
```
neuodev, line 10: Not needed.

apps/nova/src/components/TutorSettings/ProfileSettings.tsx
```diff
@@ -1,24 +1,88 @@
...
+      setPersonalInfo((prev) => ({
+        ...prev,
+        [event.target.name]: event.target.value,
+      })),
+    []
+  );
+
+  const updateTutor = useUpdateUserPersonalInfo({ onError, onSuccess });
+
+  const mutateTutor = useCallback(() => {
```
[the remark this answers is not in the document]
mostafakamar2308: yea, `mutateTutor` is a weird non-descriptive name.
I think `saveButtonHandler` doesn't reflect what we are saving, so I was considering
`saveTutorChanges` or `updateTutorInfo`
WDYT?

apps/nova/src/components/TutorSettings/TutorPersonalInfoSettings.tsx
```diff
@@ -0,0 +1,97 @@
...
+              element="caption"
+              weight="semibold"
+              className="text-brand-700"
+            >
+              {intl("global.labels.edit")}
+            </Typography>
+            <Edit className="[&>*]:stroke-brand-700" />
+          </div>
+        </div>
+        <div>TOPICS</div>
```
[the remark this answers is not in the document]
mostafakamar2308: Thank you for your hard work <3

apps/nova/src/components/TutorSettings/ProfileSettings.tsx
```diff
@@ -1,24 +1,88 @@
...
   const tutorInfo = useFindTutorInfo(id);
+  const { refetch } = useUserContext();
   const intl = useFormatMessage();
+  const toast = useToast();
+
+  const [personalInfo, setPersonalInfo] = useState<PersonalInfo>({
+    name: tutorInfo.data?.name || "",
+    bio: tutorInfo.data?.bio || "",
+    about: tutorInfo.data?.about || "",
+  });
```
neuodev, line 30: Use `react-hook-form` for validation and constancy.

apps/nova/src/components/TutorSettings/ProfileSettings.tsx
```diff
@@ -1,24 +1,88 @@
...
+
+  const onSuccess = useCallback(() => {
+    tutorInfo.refetch();
+    refetch.user();
+  }, [tutorInfo, refetch]);
+
+  const onError = useCallback(
+    (error: Error) => {
+      toast.error({
+        title: intl("profile.update.error"),
```
neuodev, line 40:
```suggestion
        title: intl("tutor-settings.profile.update.error"),
```

apps/nova/src/components/Common/Animate.tsx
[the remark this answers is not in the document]
mostafakamar2308: also the animate in Content page is defined locally, not the one I am importing. (that's why the pipeline didn't fail)
I will change this also

neuodev (changes requested):
All intl messages should go [here](https://github.com/litespace-org/litespace/blob/master/packages/luna/src/locales/ar-eg.json#L70)

packages/luna/src/locales/ar-eg.json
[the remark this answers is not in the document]
mmoehabb: Aha got it. Thanks.

apps/nova/src/components/TutorSettings/ProfileSettings.tsx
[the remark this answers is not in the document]
mmoehabb: Good point! We may combine both names, that you proposed: `saveTutorInfoChanges` :smile:

apps/nova/src/components/TutorSettings/ProfileSettings.tsx
mostafakamar2308: replied on telegram*

apps/nova/src/components/Common/Animate.tsx
mostafakamar2308: this warning
image
the main problem is that key is a reserved prop name for react, so I changed it

apps/nova/src/components/TutorSettings/ProfileSettings.tsx
mostafakamar2308: I didn't do this because I don't like to add existing words again with just different key, but I'll do it

apps/nova/src/components/Common/Animate.tsx
mostafakamar2308: and we especially need to define a key, to let react know that the element has changed and let the animation work.
if we didn't, react will cache the div and only rerender its content, thus doesn't execute the animation.
@neuodev

neuodev:
🚀 [Nova](https://nova-dws6sszvd-litespace-core.vercel.app) 
🚀 [Luna](https://luna-cb31vr86u-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-d5ysj31or-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-5tvfz26bd-litespace-core.vercel.app)

apps/nova/src/components/TutorSettings/ProfileSettings.tsx
```diff
@@ -1,24 +1,112 @@
...
+        title: intl("tutor-settings.profile.update.error"),
+        description: error.message,
+      });
+    },
+    [intl, toast]
+  );
+
+  const updateTutor = useUpdateUserPersonalInfo({
+    onError,
+    onSuccess: refetchTutorInfo,
```
mmoehabb, line 57: We may also show to users that the info has been updated successfully, by using toast, for better user experience.

mmoehabb (changes requested):
Thanks for this incredible work!

packages/luna/src/components/TopicSelectionDialog/TopicSelectionDialog.tsx
```diff
@@ -117,7 +115,9 @@ export const TopicSelectionDialog: React.FC<Props> = ({
               <LoadingError
                 size="medium"
                 error={intl("tutor-settings.topics.selection-dialog.error")}
-                retry={retry}
+                retry={() =>
+                  confirm(keys(selectionMap).map((key) => Number(key)))
+                }
               />
```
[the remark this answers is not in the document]
mostafakamar2308: I thought that is for retrying sending,
The data is fetched in another component then passed to the dialog. but you make a lot of sense.

apps/nova/src/components/TutorSettings/ProfileSettings.tsx
mostafakamar2308: in the settings page -for the student-, I was told to not do this.
WDYT?

neuodev: yup, no need to do this. User can just see his updated info reflected on the screen.

apps/nova/src/components/TutorSettings/ProfileSettings.tsx
```diff
@@ -1,24 +1,112 @@
...
+  });
+
+  const name = form.watch("name");
+  const bio = form.watch("bio");
+  const about = form.watch("about");
+
+  const refetchTutorInfo = useCallback(() => {
+    invalidateQuery([QueryKey.FindTutorInfo, tutorInfo.id]);
+    userTopics.refetch();
+    refetch.user();
```
neuodev, line 42:
```suggestion
    refetch.user();
    refetch.meta();
```

apps/nova/src/components/TutorSettings/ProfileSettings.tsx
```diff
@@ -1,24 +1,112 @@
...
+  }, [tutorInfo, invalidateQuery, refetch, userTopics]);
```
neuodev, line 43: I believe we don't need to call `.refetch` explicitly, you can just call `invalidate` with their query keys without importing their hooks.

neuodev: This should do the same thing as `.refetch()`

```suggestion
  const refetchTutorInfo = useCallback(() => {
    invalidateQuery([QueryKey.FindTutorInfo, QueryKey.FindUserInfo, QueryKey.FindUserTopics]);
  }, [tutorInfo, invalidateQuery, refetch, userTopics]);
```

apps/nova/src/components/TutorSettings/ProfileSettings.tsx
```diff
@@ -1,24 +1,112 @@
...
+
+  if (!tutorInfo) return null;
+
   return (
-    <div className="flex flex-col gap-10">
-      <div className="flex justify-between p-10 pb-0">
-        <TutorProfileCard variant="small" {...tutorInfo.data} />
-        <Button size={ButtonSize.Small}>{intl("settings.save")}</Button>
+    <Form
+      onSubmit={form.handleSubmit(saveTutorInfoChanges)}
```
neuodev, line 86:
```suggestion
  const onSbumit = useCallback(
    () => handleSubmit((data: ITutorSettingsForm) => {
      return updateTutor.mutate({
        id: tutorInfo.id,
        payload: {
          name: data.name,
          bio: data.bio,
          about: data.about,
        },
      });
    }),
    [tutorInfo.id, updateTutor]
  );

  if (!tutorInfo) return null;

  return (
    <Form
      onSubmit={onSbumit}
```

apps/nova/src/components/TutorSettings/common.ts
neuodev, line 1: We can call this file `types.ts`.

neuodev:
The [loading message is not correct](https://github.com/litespace-org/litespace/blob/d5794b63b5cd5cb2437c644d729485c3f34a07ed/apps/nova/src/pages/TutorSettings.tsx#L36).

neuodev:
[The error message is not correct.](https://github.com/litespace-org/litespace/blob/d5794b63b5cd5cb2437c644d729485c3f34a07ed/apps/nova/src/pages/TutorSettings.tsx#L46)

neuodev:
> The [loading message is not correct](https://github.com/litespace-org/litespace/blob/d5794b63b5cd5cb2437c644d729485c3f34a07ed/apps/nova/src/pages/TutorSettings.tsx#L36).

Also, the styling is not the same [as in the design](https://www.figma.com/design/4SwhwulDQ3eEDqevWprW9E/LiteSpace?node-id=5879-127188&t=Ccl74NuGJFynFyN3-4). There is no wrapping card around the spinner (or the error). @mostafakamar2308

packages/headless/src/user.ts
```diff
@@ -90,6 +90,50 @@ export function useUpdateUser({
   });
 }
 
+export function useUpdateUserPersonalInfo({
```
neuodev, line 93: Let's call this hook `useUpdateUser` becuase it just calling `user.update`. The hook with the same name we can call it `useUpdateFullUser` becuase it his multiple endpoints. 

```suggestion
export function useUpdateUser({
```

packages/headless/src/user.ts
```diff
@@ -90,6 +90,50 @@ export function useUpdateUser({
...
+
+  return useMutation({
+    mutationFn: update,
+    mutationKey: [MutationKey.UpdateUserPersonalInfo],
+    onSuccess,
+    onError,
+  });
+}
+
+export function useUpdateTutorTopics({
```
neuodev, line 115: Topics are not specific for tutors. Students can also update their topics selection. This why we should call this hook `useUpdateUserTopics`. 

```suggestion
export function useUpdateUserTopics({
```

packages/headless/src/user.ts
```diff
@@ -90,6 +90,50 @@ export function useUpdateUser({
...
+  onSuccess,
+  onError,
+}: {
+  onSuccess?: OnSuccess<void>;
+  onError?: OnError;
+}) {
+  const atlas = useAtlas();
+  const update = useCallback(
+    async ({ payload }: { payload: ITopic.ReplaceUserTopicsApiPayload }) =>
```
neuodev, line 124: No need to add an param-object. 

```suggestion
    async (payload:  payload: ITopic.ReplaceUserTopicsApiPayload) =>
```

packages/headless/src/user.ts
```diff
@@ -90,6 +90,50 @@ export function useUpdateUser({
...
+}
+
+export function useUpdateTutorTopics({
+  onSuccess,
+  onError,
+}: {
+  onSuccess?: OnSuccess<void>;
+  onError?: OnError;
+}) {
+  const atlas = useAtlas();
```
neuodev, line 122: Empty line. 

```suggestion
  const atlas = useAtlas();

```

packages/headless/src/user.ts
```diff
@@ -90,6 +90,50 @@ export function useUpdateUser({
...
 }
 
+export function useUpdateUserPersonalInfo({
+  onSuccess,
+  onError,
+}: {
+  onSuccess?: OnSuccess<IUser.Self>;
+  onError?: OnError;
+}) {
+  const atlas = useAtlas();
```
neuodev, line 100: Empty line. 

```suggestion
  const atlas = useAtlas();

```

packages/luna/src/components/TopicSelectionDialog/TopicSelectionDialog.tsx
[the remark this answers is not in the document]
neuodev:
```suggestion
              <LoadingError
                size="medium"
                error={intl("tutor-settings.topics.selection-dialog.error")}
                retry={retry}
              />
```

neuodev:
@mostafakamar2308 @mmoehabb the changes below are working as expected and doesn't cause any react or framer-motion issues.

- Removed the `name` prop. 
- Use the `key` at the `Animate` component (the key prop is not accessed by the `Animate` component).

```ts
      <AnimatePresence initial={false} mode="wait">
        {tab === "profile" ? (
          <Animate key="profile">
            <TutorPersonalInfoSettings tutor={tutor} form={form} />
          </Animate>
        ) : null}
        {tab === "settings" ? (
          <Animate key="settings">
            <div className="p-10">Settings</div>
          </Animate>
        ) : null}
      </AnimatePresence>
```


<details><summary>Animate</summary>
<p>

```tx
import React from "react";
import { motion } from "framer-motion";

export const Animate: React.FC<{ children: React.ReactNode }> = ({
  children,
}) => {
  return (
    <motion.div
      initial={{
        opacity: 0,
      }}
      animate={{
        opacity: 1,
        transition: {
          duration: 0.1,
          ease: "linear",
        },
      }}
      exit={{
        opacity: 0,
      }}
    >
      {children}
    </motion.div>
  );
};

```

</p>
</details>

apps/nova/src/components/TutorSettings/TutorPersonalInfoSettings.tsx
```diff
@@ -0,0 +1,183 @@
...
+        className="text-natural-950"
+      >
+        {intl("tutor-settings.personal-info.title")}
+      </Typography>
+      <div className="flex items-center gap-8">
+        <div className="grow flex flex-col">
+          <Label>{intl("tutor-settings.personal-info.name")}</Label>
+          <Controller.Input
+            value={tutor.name || ""}
+            control={form.control}
```
neuodev, line 101: This is not right, the value should come from the from (the same applies to other fields). 

```suggestion
            value={form.watch('name') || ""}
```

apps/nova/src/components/TutorSettings/TutorPersonalInfoSettings.tsx
```diff
@@ -0,0 +1,183 @@
...
+            rules={{ validate: validateUserName }}
+            autoComplete="off"
+            name="name"
+          />
+        </div>
+        <div className="grow flex flex-col">
+          <Label>{intl("tutor-settings.personal-info.bio")}</Label>
+          <Controller.Input
+            value={tutor.bio || ""}
```
neuodev, line 110:
```suggestion
            value={form.watch('bio')}
```

neuodev:
The form validation doesn't work @mostafakamar2308

apps/nova/src/components/TutorSettings/TutorPersonalInfoSettings.tsx
```diff
@@ -0,0 +1,183 @@
...
+import { useInvalidateQuery } from "@litespace/headless/query";
+import { QueryKey } from "@litespace/headless/constants";
+
+export const TutorPersonalInfoSettings: React.FC<{
+  tutor: {
+    video: string | null;
+    name: string | null;
+    bio: string | null;
+    about: string | null;
+    topics: { id: number; label: string }[];
```
neuodev, line 23: You don't need `tutor.topics`, you can use the one you are getting from the `useFidnUserTopics`.

neuodev (changes requested):

apps/nova/src/components/TutorSettings/TutorPersonalInfoSettings.tsx
mostafakamar2308: I am doing it [here](https://github.com/litespace-org/litespace/blob/d5794b63b5cd5cb2437c644d729485c3f34a07ed/apps/nova/src/components/TutorSettings/ProfileSettings.tsx#L107), I think it's better to group every detail about the user in the same place

apps/nova/src/components/TutorSettings/ProfileSettings.tsx
mostafakamar2308: This would caused a very annoying problem of reloading, because the FormEventHandler doesn't get passed so we can't prevent reloading by `e.preventDefault()`

neuodev:
🚀 [Nova](https://nova-6ti924vh9-litespace-core.vercel.app) 
🚀 [Luna](https://luna-oonf4yiek-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-5cd1maim2-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-r7n0m1mng-litespace-core.vercel.app)

neuodev (approved):

neuodev:
🚀 [Nova](https://nova-bliqa8rrr-litespace-core.vercel.app) 
🚀 [Luna](https://luna-f8v9rkbrs-litespace-core.vercel.app) 
🚀 [Nebula](https://nebula-kc2684xy7-litespace-core.vercel.app) 
🚀 [Dashboard](https://dashboard-r7stpoek0-litespace-core.vercel.app)

PR #499, opened 2025-03-13
added(server/tests): upload and drop assets unit test has been added. And user tests have been audited.

mmoehabb:
### Quality Checklist
- A unit test, that uploads and drops user/tutor assets, has been added.
- All `user.test.ts` unit tests has been modified to be no longer skipped. 

[ClickUp](https://app.clickup.com/t/86985cmay)

### DIFF
--- a/services/server/fixtures/mockApi.ts
+++ b/services/server/fixtures/mockApi.ts
@@ -49,6 +49,12 @@ export function mockApi<Body = object, Params = object, Query = object>(
   return async <Response>(
     request: MockRequest<Body, Params, Query>
   ): Promise<{ status: number | null; body: Response | null }> => {
+    // these assigns avoids absurd errors while using
+    // the mockApi in test suites
+    if (!request.body) request.body = {} as Body;
+    if (!request.query) request.query = {} as Query;
+    if (!request.params) request.params = {} as Params;
+
     if (typeof request.user === "number")
       request.user = await db.user({ role: request.user });
 
--- a/services/server/fixtures/mockApi.ts
+++ b/services/server/fixtures/mockApi.ts
@@ -1,4 +1,4 @@
-import Express from "express";
+import * as ExpressLib from "express";
 import { IUser } from "@litespace/types";
 import db from "@fixtures/db";
 import { ApiContext } from "@/types/api";
@@ -9,6 +9,7 @@ type MockRequest<Body = object, Params = object, Query = object> = {
   params?: Params;
   query?: Query;
   user?: IUser.Self | IUser.Role;
+  files?: Record<string, Express.Multer.File[]> | Express.Multer.File[];
 };
 
 class MockResponse<T> {
@@ -40,9 +41,9 @@ class MockResponse<T> {
 
 export function mockApi<Body = object, Params = object, Query = object>(
   handler: (
-    req: Express.Request,
-    res: Express.Response,
-    next: Express.NextFunction
+    req: ExpressLib.Request,
+    res: ExpressLib.Response,
+    next: ExpressLib.NextFunction
   ) => Promise<void> | void
 ) {
   return async <Response>(
@@ -54,8 +55,8 @@ export function mockApi<Body = object, Params = object, Query = object>(
     const response = new MockResponse<Response>();
     return await new Promise((resolve) => {
       const result = handler(
-        request as unknown as Express.Request,
-        response as unknown as Express.Response,
+        request as unknown as ExpressLib.Request,
+        response as unknown as ExpressLib.Response,
         (next) => resolve(next)
       );
 
--- a/services/server/src/handlers/user.ts
+++ b/services/server/src/handlers/user.ts
@@ -44,7 +44,6 @@ import {
   joinTutorCache,
   orderTutors,
 } from "@/lib/tutor";
-import { ApiContext } from "@/types/api";
 import { asIsoDate } from "@litespace/utils/dayjs";
 import {
   isAdmin,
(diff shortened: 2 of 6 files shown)


services/server/fixtures/mockApi.ts
```diff
@@ -1,4 +1,4 @@
-import Express from "express";
+import * as ExpressLib from "express";
```
neuodev, line 1: I belive there is no need for this. This pattern is not common in our codebase.

```suggestion
import {Request, Response, NextFunction} from "express";
```

services/server/tests/handlers/user.test.ts
```diff
@@ -12,19 +10,49 @@ import { Role } from "@litespace/types/dist/esm/user";
...
+  fieldname: "mockfield",
+  originalname: "mockname",
+  encoding: "testing",
+  mimetype: "testing",
+  size: 1,
+  stream: new Readable(),
+  destination: "empty",
+  path: "/tmp",
+  buffer: Buffer.from(new ArrayBuffer()),
+});
```
neuodev, line 45: Super nice!

services/server/src/handlers/user.ts
```diff
@@ -189,122 +188,116 @@ export async function create(req: Request, res: Response, next: NextFunction) {
...
+    bio,
+    about,
+    notice,
+    phone,
+    city,
+  }: IUser.UpdateApiPayload = updateUserPayload.parse(req.body);
+
+  // return forbidden if the currentUser is neither admin nor studio and tring to update other user data
+  if (currentUser.id !== targetUser.id && isRegularUser(currentUser))
+    return next(forbidden());
```
neuodev, line 221: This is an important check! 

----

* Instead of describing the code, let's describe the "rules". 
* Let's make the code self descriptive.

```suggestion
  // It is forbidden for regular users to update other users.
  const updatingSelf = currentUser.id === targetUser.id
  if (!updatingSelf && isRegularUser(currentUser)) return next(forbidden());
```

services/server/src/handlers/user.ts
```diff
@@ -189,122 +188,116 @@ export async function create(req: Request, res: Response, next: NextFunction) {
...
+
+  // return forbidden if the studio is trying to drop user image
+  if (isStudio(currentUser) && image === null) return next(forbidden());
```
neuodev, line 224: I don't uderstand why this is not allowed. Studios can remove tutor images & videos. 

I belive the logic should look something like this: 

```suggestion
  // Studios are not allowed to update other users other than tutors (including tutor managers)
  // Studios can only update image, video, and thumbnail and nothing else.
  const eligibleStudio =isStudio(currentUser) && isTutor(targetUser)  && targetUser.studioId === currentUser.id && (image === null || video === null || thmbnaill === null)
  if (isStudio(currentUser) && !updatingTutor) return next(forbidden());
```

> [!NOTE]
> Let's add unit tests for this.

services/server/src/handlers/user.ts
```diff
@@ -189,122 +188,116 @@ export async function create(req: Request, res: Response, next: NextFunction) {
...
+
+  // return forbidden if the studio is trying to drop media for an unassociated user
+  if (isStudio(currentUser) && targetTutor?.studioId !== currentUser.id)
+    return next(forbidden());
```
neuodev, line 228: Let's combain this check with the check above.

services/server/src/handlers/user.ts
```diff
@@ -189,122 +188,116 @@ export async function create(req: Request, res: Response, next: NextFunction) {
...
-            );
-
-          return user;
-        }
+    const isValidCurrentPassword =
+      (password.current === null && expectedPasswordHash === null) ||
+      (password.current &&
+        expectedPasswordHash &&
+        isSamePassword(password.current, expectedPasswordHash));
+    if (!isValidCurrentPassword) return next(wrongPassword());
```
neuodev, line 240: Let's simplify the code here.

```suggestion
  const noPassword = password.current === null && expectedPasswordHash === null
  const samePassword = password.current &&
        expectedPasswordHash &&
        isSamePassword(password.current, expectedPasswordHash)
    const isValidCurrentPassword = noPassword || samePassword;
    if (!isValidCurrentPassword) return next(wrongPassword());
```

services/server/src/handlers/user.ts
```diff
@@ -189,122 +188,116 @@ export async function create(req: Request, res: Response, next: NextFunction) {
...
+          name,
+          email,
+          gender,
+          birthYear,
+          phone,
+          image,
+          // Reset user verification status incase his email updated.
+          verifiedEmail: email ? false : undefined,
+          password: password ? hashPassword(password.new) : undefined,
+        };
```
neuodev, line 268: Why are we doing this check? Why if the current user is a studio we are not updating any thing?

neuodev (changes requested):

services/server/fixtures/mockApi.ts
mmoehabb: I have encountered an absurd TS error. This workaround allows me to use the generic functionality in the return value of the mockApi function:

I have no clue how to solve the problem otherwise.
```TS
export function mockApi<Body = object, Params = object, Query = object>(
  handler: (
    req: ExpressLib.Request,
    res: ExpressLib.Response,
    next: ExpressLib.NextFunction
  ) => Promise<void> | void
) {
  return async <Response>(
    request: MockRequest<Body, Params, Query>
  ): Promise<{ status: number | null; body: Response | null }> => {
...
```

services/server/src/handlers/user.ts
mmoehabb: If my memory doesn't fail me, the specification was that studios can only update tutors data (thumbnail & video). With the current specification this part would be as follow:
```TS
    const updatePayload = isStudio(currentUser)
      ? {
        image,
      }
      : {
          city,
          name,
          gender,
          birthYear,
          phone,
          image,
          email,
          // Reset user verification status incase his email updated.
          verifiedEmail: email ? false : undefined,
          password: password ? hashPassword(password.new) : undefined,
          enabledTelegram,
          enabledWhatsapp,
        };
```

neuodev:
I am closing this pull request becuase it is out of sync with the latest master and there is a lot of conflicts.

PR #706, opened 2025-05-16
impl(api/models): reports model and api initial implementation.

mmoehabb:
### Quality Checklist

- [x] I ensured that the this pull request satisfies all the requirements in the associated task.
- [x] I performed a self-review.

---

[ClickUp](https://app.clickup.com/t/86990rgc2)


packages/atlas/src/api/report.ts
```diff
@@ -1,35 +1,22 @@
 import { Base } from "@/lib/base";
-import { IReport } from "@litespace/types";
+import { IReport, Paginated } from "@litespace/types";
 
 export class Report extends Base {
-  async create(payload: IReport.CreateApiPayload): Promise<IReport.Self> {
+  async create(payload: IReport.CreateApiPayload): Promise<void> {
     return this.client
-      .post<IReport.Self>(`/api/v1/report`, JSON.stringify(payload))
+      .post(`/api/v1/report`, JSON.stringify(payload))
```
mostafakamar2308, line 7: I think we have different method to define api calls like in `plans.ts`
```
  async create(payload: IPlan.CreateApiPayload): Promise<void> {
    await this.post({ route: "/api/v1/plan", payload });
  }
```

packages/models/fixtures/db.ts
```diff
@@ -41,6 +43,7 @@ import { percentage, price } from "@litespace/utils";
 
 export async function flush() {
   await knex.transaction(async (tx) => {
+    await reports.builder(tx).del();
```
mostafakamar2308, line 46: Very Nice 🚀

packages/models/migrations/1747300348202_reports.ts
```diff
@@ -0,0 +1,22 @@
+import { MigrationBuilder } from "node-pg-migrate";
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.createTable("reports", {
+    id: { type: "serial", primaryKey: true, notNull: true },
+    user_id: { type: "int", references: "users(id)", default: null },
+    title: { type: "VARCHAR(128)", notNull: true },
+    description: { type: "VARCHAR(255)", notNull: true },
```
mostafakamar2308, line 8: I think people may want to write more in the description, wdyt?

packages/models/src/query.ts
```diff
@@ -189,10 +189,10 @@ export function withBooleanFilter<R extends object, T>(
 export function withNullableFilter<R extends object, T>(
   builder: Knex.QueryBuilder<R, T>,
   column: string,
-  value?: boolean
+  notnull?: boolean
 ): Knex.QueryBuilder<R, T> {
-  if (typeof value === "undefined") return builder;
-  return builder.where(column, value ? "IS NOT" : "IS", null);
+  if (typeof notnull === "undefined") return builder;
```
mostafakamar2308, line 194: we may name it `notNull`

packages/models/src/reports.ts
```diff
@@ -1,99 +1,134 @@
...
     if (!row) throw new Error("Report not found; should never happen");
     return this.from(row);
   }
 
   async update(
     id: number,
-    payload: IReport.UpdatePayload
+    payload: WithOptionalTx<IReport.UpdatePayload>
   ): Promise<IReport.Self> {
+    const now = new Date();
```
mostafakamar2308, line 46: It's better to use `dayjs` and also in `utc` time

packages/models/src/reports.ts
```diff
@@ -1,99 +1,134 @@
...
+    withNullableFilter(builder, this.column("screenshot"), screenshot);
+    withNullableFilter(builder, this.column("log"), log);
+
+    // ============== date fields ========
+    withDateFilter(builder, this.column("created_at"), createdAt);
+    withDateFilter(builder, this.column("updated_at"), updatedAt);
+
+    // ==============  list-based fileds ========
+    withListFilter(builder, this.column("id"), ids);
+    withListFilter(builder, this.column("user_id"), users);
```
mostafakamar2308, line 93: Very Clean indeed 😄

packages/types/src/report.ts
```diff
@@ -1,59 +1,65 @@
...
-  };
+export type UpdatePayload = {
+  resolved: boolean;
+};
 
-export type MappedAttributes = SelectableSelf &
-  MappedRecordAttributes & {
-    resolvedAt: string | null;
-  };
+export type ModelFindQuery = IFilter.SkippablePagination & {
```
mostafakamar2308, line 40: I think the convention is putting the package name last like `UpdateRoomApiPayload`.
So I suggest something like `FindQueryModelPayload`

packages/ui/src/constants/error.ts
```diff
@@ -33,7 +33,6 @@ export const apiErrorMap: Record<ApiErrorCode, LocalId> = {
   [ApiError.InterviewNotFound]: "error.api.interview-not-found",
   [ApiError.TopicNotFound]: "error.api.topic-not-found",
   [ApiError.InterviewAlreadySigned]: "error.api.interview-already-signed",
-  [ApiError.ReportReplyNotFound]: "error.api.report-reply-not-found",
```
mostafakamar2308, line 36: Nice attention to small details 🚀

services/server/src/handlers/report.ts
```diff
@@ -0,0 +1,99 @@
...
+    createReportPayload.parse(req.body);
+
+  await reports.create({
+    userId: user.id,
+    title,
+    description,
+    log,
+  });
+
+  res.status(200);
```
mostafakamar2308, line 60: In case of handlers that doesn't return data to the client It's better to use `sendStatus` as it sets and sends the status. so the req will be closed.

see [this](https://stackoverflow.com/questions/38620682/difference-between-response-status-vs-response-sendstatus-in-express) answer for further details

services/server/src/handlers/report.ts
```diff
@@ -0,0 +1,99 @@
...
+  page: zod.optional(pageNumber).default(paginationDefaults.page),
+  size: zod.optional(pageSize).default(paginationDefaults.size),
+});
+
+async function create(req: Request, res: Response, next: NextFunction) {
+  const user = req.user;
+  const allowed = isUser(user);
+  if (!allowed) return next(forbidden());
+
+  const { title, description, log }: IReport.CreateApiPayload =
```
mostafakamar2308, line 50: what about uploading the screenshot? also isn't the log another file that would be uploaded?

mostafakamar2308 (changes requested):
Great work as always, I just left some inquiries and some minor suggestions. but overall clean structured logic as always ❤️

neuodev:
@mmoehabb please resolve the conflicts and address the comments from @mostafakamar2308

packages/types/src/report.ts
mmoehabb: Or maybe `FindModelQuery`.
The patterns as follows: `{method}{scope}{struct-type}`
Examples: `FindApiQuery`, `FindModelQuery`, `CreateApiPayload`, `CreateModelPayload` (the convention currently is to drop the model word in the middle),... etc.

services/server/src/handlers/report.ts
mmoehabb: I have no idea how am I dropped that. Maybe forgot to push the last changes. Anyway, checkout it out now, and tell me what do you think.

services/server/src/handlers/report.ts
```diff
@@ -0,0 +1,120 @@
...
+  title: string.optional(),
+  description: string.optional(),
+  screenshot: boolean.optional(),
+  log: boolean.optional(),
+  resolved: boolean.optional(),
+  createdAt: dateFilter.optional(),
+  updatedAt: dateFilter.optional(),
+  page: zod.optional(pageNumber).default(paginationDefaults.page),
+  size: zod.optional(pageSize).default(paginationDefaults.size),
+});
```
neuodev, line 45: The description for each field is missing.

services/server/src/handlers/report.ts
```diff
@@ -0,0 +1,120 @@
...
+  users: zod.array(number).optional(),
+  title: string.optional(),
+  description: string.optional(),
+  screenshot: boolean.optional(),
+  log: boolean.optional(),
+  resolved: boolean.optional(),
+  createdAt: dateFilter.optional(),
+  updatedAt: dateFilter.optional(),
+  page: zod.optional(pageNumber).default(paginationDefaults.page),
+  size: zod.optional(pageSize).default(paginationDefaults.size),
```
neuodev, line 44:
```suggestion
  page: pageNumber.optional().default(paginationDefaults.page),
  size: pageSize.optional().default(paginationDefaults.size),
```

packages/types/src/report.ts
```diff
@@ -1,59 +1,68 @@
...
+  log?: string;
 };
 
-export type UpdatePayload = Partial<
-  Omit<CreatePayload, "createdBy"> & { resolved: boolean }
-> & {
-  updatedBy: number;
+export type UpdateApiPayload = {
+  resolved: boolean;
 };
```
neuodev, line 65: I was thinking about this for some time now and here is my thoughts about it. 

## The common wisdom 

The convention is to include the resource that is about to be updated in the the url/route. For example: `/report/:id`. The client will then replace `:id` with the actual resource id and actuall data that should be updated will be in the body (or some time the query). This how the most update-api routes are implemented. 

## Drawbacks 

Once of the core concepts that our codebase is built around is to achive the maximum type safety possible. This way of providing the param is error prone. For example: chaning the api route schema will cause the request to fail and the client must be updated. Also, the client can pass any data of any type for example it can pass a string isntead of a number or anything else. 

## Proposed solution 

My proposal is to include the resourse `id` in the request body and not in the url/route. This way we can achieve type safety and both the client and server will be tightly coupled. 

Example: 

```ts
type UpdateApiPayload = {
    id: number,
    activated?: boolean;
    name?: string;
}
```

Note that the id is required but the other params are optional. Note that if the the update payload has only one updatable property, it makes sense to make it required. 

```ts
type UpdateApiPayload = {
    id: number,
    resolved: boolean;
}
```

cc @mmoehabb @mostafakamar2308

packages/types/src/report.ts
```diff
@@ -1,59 +1,68 @@
...
+export type CreateApiPayload = {
   title: string;
   description: string;
-  category: string;
-  createdBy: number;
+  /**
+   * An image encoded in base64
+   */
+  screenshot?: string;
+  log?: string;
```
neuodev, line 59: The image should be uploaded as a file `blob` or a `buffer`. This mean that the screenshot or the log file will not be included in body isntead will be included as a form file `req.files`. Check how the tutor assets are being uploaded.

services/server/src/handlers/report.ts
```diff
@@ -0,0 +1,120 @@
...
+import { paginationDefaults } from "@/constants";
+import { IReport } from "@litespace/types";
+import { upload } from "@/lib/assets";
+import { v4 as uuid } from "uuid";
+
+const createReportPayload = zod.object({
+  title: string,
+  description: string,
+  screenshot: string.startsWith("data:image/").optional(),
+  log: string.optional(),
```
neuodev, line 26:
```suggestion
```

services/server/src/routes/report.ts
```diff
@@ -0,0 +1,10 @@
+import report from "@/handlers/report";
+import { Router } from "express";
+
+const router = Router();
+
+router.get("/list", report.find);
+router.post("/", report.create);
```
neuodev, line 7: Let's follow the same pattern used [here](https://github.com/litespace-org/litespace/blob/master/services/server/src/routes/user.ts#L21-L25).

services/server/src/lib/assets.ts
```diff
@@ -19,13 +19,15 @@ export async function upload({
...
   data: Buffer;
   type?: string;
   key?: string | null;
+  prefix?: string;
 }) {
   const id = key || uuid();
-  await s3.put({ key: id, data, type });
+  await s3.put({ key: `${prefix}${id}`, data, type });
   return id;
 }
```
neuodev, line 32: This will cause a serious issue.  Can you spot the issue? 

@mmoehabb @mostafakamar2308

services/server/src/handlers/report.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+async function find(req: Request, res: Response, next: NextFunction) {
+  const user = req.user;
+  const allowed = isAdmin(user);
+  if (!allowed) return next(forbidden());
+
+  const payload: IReport.FindApiPayload = findReportsPayload.parse(req.body);
+
+  const { list, total } = await reports.find(payload);
```
neuodev, line 109: We cannot pass the `payload` directlly to `reports.find` as an hacker can pass extra params (if he has knowledge of the codebase or not) that is not included in the schema but can be provided to `reports.find` (think about it or try it!). it is better to destruct the properties you are intested in or make sure that the payload doesn't have any unlegal params. 


cc @mostafakamar2308

services/server/tests/handlers/report.test.ts
```diff
@@ -0,0 +1,158 @@
...
+  handlers.update
+);
+const create = mockApi<IReport.CreateApiPayload>(handlers.create);
+
+describe("/api/v1/report/", () => {
+  beforeEach(async () => {
+    await db.flush();
+  });
+
+  describe("GET /api/v1/report/list", () => {
```
neuodev, line 26: Let's use the name of the function istead of the route.

neuodev (changes requested):

services/server/src/lib/assets.ts
mmoehabb: Nope. It's not obvious to me. Do you mean with the commit modification or the function implementation overall?

neuodev:
🚀 [WEB](https://web-ctjxv6scz-litespace-core.vercel.app) 
🚀 [UI](https://ui-hfbjqsy8h-litespace-core.vercel.app) 
🚀 [LANDING](https://landing-4ipl5r81a-litespace-core.vercel.app) 
🚀 [DASHBOARD](https://dashboard-7vigcxx9h-litespace-core.vercel.app)

packages/models/src/reports.ts
```diff
@@ -1,99 +1,146 @@
...
-          category: payload.category,
-          resolved: payload.resolved,
-          resolved_at: payload.resolved ? new Date() : undefined,
-          updated_at: new Date(),
-          updated_by: payload.updatedBy,
-        },
-        "*"
-      )
+    payload: WithOptionalTx<IReport.UpdateModelPayload>
+  ): Promise<void> {
```
neuodev, line 46:
```
async update(payload: WithOptionalTx<IReportUpdatePayload>): Promise<Void>;
```

Based on the new convension, let's include the id as part of the payload.

packages/models/src/reports.ts
```diff
@@ -1,99 +1,146 @@
...
+    const now = dayjs().utc();
+    await knex("reports")
```
neuodev, line 48:
```suggestion
    await this.builder(tx)
```

packages/models/src/reports.ts
```diff
@@ -1,99 +1,146 @@
...
+      .update({
+        resolved: payload.resolved,
+        updated_at: now.toDate(),
+      })
       .where("id", id);
```
neuodev, line 53:
```suggestion
      .where(this.column("id"), id);
```

packages/models/src/reports.ts
```diff
@@ -1,99 +1,146 @@
...
+    withDateFilter(builder, this.column("updated_at"), updatedAt);
+
+    // ==============  list-based fileds ========
+    withListFilter(builder, this.column("id"), ids);
+    withListFilter(builder, this.column("user_id"), users);
+
+    const total = await countRows(builder.clone(), { distinct: true });
+    const query = builder
+      .select<IReport.Row[]>("*")
+      .orderBy(this.column("created_at"), "desc");
```
neuodev, line 93: You should filter by both the creation time and the id. Incase two records was created at the same time. 

```
      .orderBy([
        {
          column: this.column("created_at"),
          order: "desc",
        },
        {
          column: this.column("id"),
          order: "desc",
        },
      ]);
```

packages/models/src/reports.ts
```diff
@@ -1,99 +1,146 @@
...
-    }));
+  async findManyBy<T extends keyof IReport.Row>(
+    key: T,
+    value: IReport.Row[T]
+  ): Promise<IReport.Self[]> {
+    const rows = await knex<IReport.Row>(this.table)
+      .select("*")
+      .where(key, value);
+    return rows.map((row) => this.from(row));
   }
```
neuodev, line 120: We don't need these two function. All what we need is just one function called `findOne` and it should be a wrapper of `find` and returns the first item in the list check `interviews.find` and `interviews.findOne`

packages/types/src/api.ts
```diff
@@ -32,6 +32,7 @@ export enum ApiError {
   ReportNotFound = "report-not-found",
   InterviewNotFound = "interview-not-found",
   TopicNotFound = "topic-not-found",
+  InterviewAlreadySigned = "interview-already-signed",
```
neuodev, line 35: This type should be removed.

packages/ui/src/locales/ar-eg.json
```diff
@@ -975,6 +976,7 @@
   "error.api.expired-verification-code": "هذا الكود انتهت صلاحيته، برجاء المحاولة مرة أخرى",
   "error.api.invalid-verification-code": "هذا الكود خاطئ، برجاء التأكد من صحة الكود والمحاولة مرة أخرى",
   "error.api.incorrect-phone": "هذا الرقم خاطئ، برجاء التأكد من صحة الكود والمحاولة مرة أخرى",
+  "error.api.large-file-size": "عذرا، حجم الملف الذي تحاول رفعه قد تعدى السعة المسموحة.",
```
neuodev, line 979:
```suggestion
  "error.api.large-file-size": "حجم الملف الذي تحاول رفعه قد تعدى السعة المتاحة.",
```

services/server/src/handlers/report.ts
```diff
@@ -0,0 +1,140 @@
...
+  const key = uuid();
+
+  // upload the screenshot if exists
+  const screenshot = getRequestFile(
+    req.files,
+    IReport.AssetFileName.Screenshot
+  );
+  if (screenshot && exceedsSizeLimit(screenshot.size) !== false)
+    return next(largeFileSize());
+  const screenshotKey = screenshot
```
neuodev, line 62:
```suggestion
  );
  
  if (screenshot && exceedsSizeLimit(screenshot.size) !== false)
    return next(largeFileSize());
    
  const screenshotKey = screenshot
```

services/server/src/handlers/report.ts
```diff
@@ -0,0 +1,140 @@
...
+        type: screenshot.mimetype,
+        prefix: "reports/screenshots/",
+        key,
+      })
+    : undefined;
+
+  // upload the log file if exists
+  const log = getRequestFile(req.files, IReport.AssetFileName.Log);
+  if (log && exceedsSizeLimit(log.size) !== false) return next(largeFileSize());
+  const logKey = log
```
neuodev, line 74: empty lines for redablity. 
```suggestion
  // upload the log file if exists
  const log = getRequestFile(req.files, IReport.AssetFileName.Log);
  
  if (log && exceedsSizeLimit(log.size) !== false) return next(largeFileSize());
  
  const logKey = log
```

packages/models/migrations/1747300348202_reports.ts
```diff
@@ -0,0 +1,22 @@
+import { MigrationBuilder } from "node-pg-migrate";
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.createTable("reports", {
+    id: { type: "serial", primaryKey: true, notNull: true },
+    user_id: { type: "int", references: "users(id)", default: null },
+    title: { type: "varchar(128)", notNull: true },
+    description: { type: "varchar(1000)", notNull: true },
+    screenshot: { type: "varchar(64)", notNull: false },
+    log: { type: "varchar(64)", notNull: false },
```
neuodev, line 10: It should be a fixed size 32 characters `chars(32)`

services/server/src/handlers/report.ts
```diff
@@ -0,0 +1,140 @@
...
+    req.files,
+    IReport.AssetFileName.Screenshot
+  );
+  if (screenshot && exceedsSizeLimit(screenshot.size) !== false)
+    return next(largeFileSize());
+  const screenshotKey = screenshot
+    ? await upload({
+        data: screenshot.buffer,
+        type: screenshot.mimetype,
+        prefix: "reports/screenshots/",
```
neuodev, line 66: Let's remove the concepts of the prefix as it will complicates our process a little bit for example, we should store the prefix with the file or when we are fetching the file from s3 we should add the prefix again. The logic is not complete.

services/server/src/handlers/report.ts
```diff
@@ -0,0 +1,140 @@
...
+    screenshot: payload.screenshot,
+    log: payload.log,
+    resolved: payload.resolved,
+    createdAt: payload.createdAt,
+    updatedAt: payload.updatedAt,
+    page: payload.page,
+    size: payload.size,
+  });
+
+  const response: IReport.FindApiResponse = { list, total };
```
neuodev, line 132: Please note that the items you are returing to the front end doesn't include the screenshot or the log file urls instead it only include its ids, we should generate a view link for these files before responding to the client. Check the function call `withImgUrl` and `withImgUrls` to get some idea.

services/server/src/lib/assets.ts
```diff
@@ -3,6 +3,8 @@ import { Request } from "express";
 import { isArray } from "lodash";
 import { v4 as uuid } from "uuid";
 import multer from "multer";
+import bytes from "bytes";
+import { logger } from "@litespace/utils";
```
neuodev, line 7: We don't use this logger. It is deprecated. Just add console.error or throw and error

services/server/src/lib/assets.ts
```diff
@@ -19,16 +21,30 @@ export async function upload({
...
+
+/**
+ * return null in case of unexpected error
+ */
+export function exceedsSizeLimit(size: number): boolean | null {
+  const limit = bytes("8mb");
+  if (!limit) {
+    logger().error("exceedsSizeLimit: an error occured with bytes function.");
+    return null;
+  }
```
neuodev, line 48: It is fine to throw an error here. No need to return a null.

services/server/src/lib/assets.ts
```diff
@@ -19,16 +21,30 @@ export async function upload({
...
+  return size > limit;
+}
```
neuodev, line 50: Not all files will have a max size of 8mb. Let's update this function look as follow: 

```suggestion
export function exceedsSizeLimit(size: number, max: "${number}mb"): boolean | null {
  const limit = bytes("8mb");
  if (!limit) throw new Error(`invalid size: ${max}`);
  return size > limit;
}
```

services/server/tests/handlers/report.test.ts
```diff
@@ -0,0 +1,161 @@
...
+const find = mockApi<
+  IReport.FindApiPayload,
+  object,
+  object,
+  IReport.FindApiResponse
+>(handlers.find);
+const update = mockApi<IReport.UpdateApiPayload>(handlers.update);
+const create = mockApi<IReport.CreateApiPayload>(handlers.create);
+
+describe(nameof(find), () => {
```
neuodev, line 20: The first outer can be just `Reports` and then each handler should has its own describe block.

neuodev (changes requested):

packages/models/src/reports.ts
mmoehabb: I just followed the convention in the interview. Shall we modify the other scripts to follow the current description?

mmoehabb: Moreover, the current description is not clear. What's  the definition of `findOne`? Is it a generic function just like `findOneBy` that defined above? or is it a specific function that retrieves a specific row according to its id?

services/server/src/handlers/report.ts
mmoehabb: The prefix is part of the key. It's where the files are located in the (digital ocean) object storage. Prefixes are for directories and more organized storage. Checkout my comment here: [https://github.com/litespace-org/litespace/pull/706#discussion_r2138273378](https://github.com/litespace-org/litespace/pull/706#discussion_r2138273378)

neuodev:
🚀 [WEB](https://web-km201txmg-litespace-core.vercel.app) 
🚀 [UI](https://ui-3gzhd0yhj-litespace-core.vercel.app) 
🚀 [LANDING](https://landing-4ipl5r81a-litespace-core.vercel.app) 
🚀 [DASHBOARD](https://dashboard-7dejfdad9-litespace-core.vercel.app)

neuodev:
🚀 [WEB](https://web-fdp7ksaf3-litespace-core.vercel.app) 
🚀 [UI](https://ui-3h4e4imw9-litespace-core.vercel.app) 
🚀 [LANDING](https://landing-4ipl5r81a-litespace-core.vercel.app) 
🚀 [DASHBOARD](https://dashboard-dh8seh5jg-litespace-core.vercel.app)

neuodev (approved):
Great work @mmoehabb!

mostafakamar2308 (approved):

PR #1101, opened 2025-10-19
feat(web): add navbar timer.

moalidv:
### Summary

Render timer to the next lesson in alternative with subscribtion quote.
Navbar is removed from both chat page with active chat room and lesson page.

Cases that should be tested: 

- [ ] subscribed without lesson 
- [ ] subscribed with upcoming lesson before the lesson with more than three minutes 
- [ ] subscribed with upcoming lesson before the lesson with less than three minutes 
- [ ] subscribed with upcoming lesson after the lesson started 
- [ ] subscribed with upcoming lesson after the lesson ended
- [ ] unsubscribed without lesson 
- [ ] unsubscribed with upcoming lesson before the lesson with more than three minutes 
- [ ] unsubscribed with upcoming lesson before the lesson with less than three minutes 
- [ ] unsubscribed with upcoming lesson after the lesson started 
- [ ] unsubscribed with upcoming lesson after the lesson ended
- [ ] chat page
- [ ] lesson page


mmoehabb (changes requested):
- Subscribe now is not shown in the desktop screen.
- The bar doesn't look professional on relativily large phone screens, or tablets.
- The text should take all required space in large screens.
- Make the counter appears just like the design. There should be more space(s) between the digits.
- The timer is hidden for a couple seconds when the page starts loading (try to refresh the page while it's counting).
- The timer appears like this: "0 : NaN : NaN" when the session time is close.

PR #1007, opened 2025-09-16
update(web): update student settings page design

moalidv:
### Quality Checklist

- [x] I took the time to compare my implementation with the design.
- [x] I ensured that the this pull request satisfies all the requirements in the associated task.
- [x] I performed a self-review.

### Links

- [MatterMost](#task-link-here)

- [Context Video](#context-video-here)

> [!IMPORTANT]
> Incase you have any valuable knowledge that you think that it is worth writing down, feel free to add it to our handbook. Wounder why we are doing this? [check this guide](https://handbook.litespace.org/s/general/p/everything-should-have-a-url-QfllToC5EY).

PR #1122, opened 2025-10-22
impl: new settings design for mobile layout.

moalidv:
### Summary

make the settings page responsive in the new design but in the old implementation


moalidv:
The code of this PR is rewriten in another [PR](https://github.com/litespace-org/litespace/pull/1007)

PR #1138, opened 2025-10-28
update(ui): update validation of phone to be globally

moalidv:


packages/ui/src/lib/validate.ts
```diff
@@ -49,6 +49,21 @@ export function validatePhone(phone: string): LocalId | null {
   return "error.phone-number.invlaid";
 }
 
+export function validatePhoneGlobally(
+  phone: string,
+  region?: string
```
mmoehabb, line 54: Why the region is optional?
I think we should make it required in order to be able to spot the places where it must be modified allover the codebase. Moreover, it should have a more specific type rather than 'string'.

mmoehabb (changes requested):
Great work. I have just one comment please give it a look.

